PR #1, opened 2018-04-25
Install libbcc into Travis CI build

bobrik:


.travis.yml
```diff
@@ -5,6 +5,12 @@ go:
   - master
 
 before_install:
+  # Install bcc
+  - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 648A4A16A23015EEF4A66B8E4052245BD4284CDD
+  - echo "deb https://repo.iovisor.org/apt/xenial xenial main" | sudo tee /etc/apt/sources.list.d/iovisor.list
```
dqminh: travis.ci has some shortcuts for this https://docs.travis-ci.com/user/installing-dependencies/#Adding-APT-Sources

bobrik: It doesn't seem to support dynamic release name and it requires `key_url`, when I only have key id.

PR #12, opened 2018-04-29
Fixing cases for fields that contains space

dqminh:
currently we split fields using space separator. But this failed in case program has space in name. For example "tmux: server". This refactor the current decoders into simple index-based decoders to let them have greater flexibility in parsing its required data.

Before:
```
Error getting table "counts" values for metric "dcache_ops_total" of program "dcstat": key "{ 0x1 \"tmux: server\" }" has 3 elements, but we expect 2
```

After:
```
ebpf_exporter_dcache_ops_total{command="tmux: server",op="refs"} 2517
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -564,8 +564,8 @@ An example to report metrics only for `systemd-journal` and `syslog-ng`:
     - name: string
     - name: regexp
       regexps:
-        - ^systemd-journal$
-        - ^syslog-ng$
+        - systemd-journal\b
+        - syslog-ng\b
 ```
 
 #### `static_map`
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ go:
   - 1.x
   - master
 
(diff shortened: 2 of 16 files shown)


bobrik:
Would it be better to split `entry.Key` into elements outside of decoders? Keeping decoders simple and only letting them read the exact string they are supposed to read seems like a cleaner cut of responsibilities.

Keys as byte slices are coming as well: https://github.com/iovisor/gobpf/pull/102.

README.md
```diff
@@ -564,8 +564,8 @@ An example to report metrics only for `systemd-journal` and `syslog-ng`:
     - name: string
     - name: regexp
       regexps:
-        - ^systemd-journal$
-        - ^syslog-ng$
+        - systemd-journal\b
```
bobrik: Why is this changing?

bobrik:
Going with #24 and byte slice decoding instead.

PR #14, opened 2018-05-06
Add runqlat example

bobrik:
Excluding swapper threads as proposed upstream:

* https://github.com/iovisor/bcc/issues/1732

cc @dqminh


bobrik:
Two spikes are kernel rebuilds:

image

dqminh:
LGTM

PR #24, opened 2018-06-26
Use byte slices to decode keys and values in maps

bobrik:


bobrik:
Build fails, because it needs https://github.com/iovisor/gobpf/pull/130, which is not incorporated into https://github.com/iovisor/gobpf/pull/112.

PR #33, opened 2018-08-17
Export enabled programs as a metric

lmb:
This is useful to check which programs are active at any point in time.

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -17,21 +17,30 @@ const prometheusNamespace = "ebpf_exporter"
 
 // Exporter is a ebpf_exporter instance implementing prometheus.Collector
 type Exporter struct {
-	config   config.Config
-	modules  map[string]*bcc.Module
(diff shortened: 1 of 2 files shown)


exporter/exporter.go
```diff
@@ -109,8 +120,12 @@ func (e *Exporter) Describe(ch chan<- *prometheus.Desc) {
 	}
 }
 
-// Collect satisfies prometeus.Collector interface and sends all metrics
+// Collect satisfies prometheus.Collector interface and sends all metrics
 func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
+	for _, program := range e.config.Programs {
+		ch <- prometheus.MustNewConstMetric(e.enabledProgramsDesc, prometheus.Gauge, 1, program.Name)
```
bobrik: I think you want `prometheus.GaugeValue`

bobrik:
You can revert #27, it's no longer necessary.

PR #51, opened 2019-04-08
Add example of measuring accept latency for sockets

dqminh:
Duration that sockets stay in accept queue can be an indicator of
applications being slow in calling `accept` syscall (i.e. nginx event loop
or go routines are slow etc.)

By measuring it, we have some signals on how much improvements were made
with changes that should speed up the user-space applications.

Signed-off-by: Daniel Dao <dqminh@cloudflare.com>

### DIFF
--- a/examples/accept.yaml
+++ b/examples/accept.yaml
@@ -1,0 +1,96 @@
+programs:
+  - name: accept
+    metrics:
+      histograms:
+        - name: accept_latency_seconds
+          help: Latency histogram for how long sockets sit in accept queue
+          table: accept_latency
+          bucket_type: exp2
+          bucket_min: 0
+          bucket_max: 26
+          bucket_multiplier: 0.000001 # microseconds to seconds
+          labels:
+            - name: port
+              size: 8
+              decoders:
+                - name: uint
+            - name: bucket
+              size: 8
+              decoders:
+                - name: uint
+    kprobes:
+      inet_csk_reqsk_queue_add: trace_inet_csk_reqsk_queue_add_start
+      inet_csk_accept: trace_inet_csk_accept_start
+    code: |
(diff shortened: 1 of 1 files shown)


examples/accept.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+      int trace_inet_csk_reqsk_queue_add_start(struct pt_regs *ctx, struct sock *sk, struct request_sock *req)
+      {
+          u64 ts = bpf_ktime_get_ns();
+          start.update(&req, &ts);
+          return 0;
+      }
+
+      int trace_inet_csk_accept_start(struct pt_regs *ctx, struct sock *sk)
+      {
```
bobrik: Other examples keep `{` at the same line with the function signature, let's keep it that way for consistency.

examples/accept.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+          delta = bpf_ktime_get_ns() - *tsp;
+          delta /= 1000;
+
+          // Latency histogram key
+          u64 latency_slot = bpf_log2l(delta);
+
+          // Cap latency bucket at max value
+          if (latency_slot > max_latency_slot) {
+              latency_slot = max_latency_slot; }
```
bobrik: Let's put closing `}` on the next line

examples/accept.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+      typedef struct listen_socket_key {
+          u16 port;
+          u64 slot;
+      } listen_socket_t;
+
+      // 27 buckets for latency, max range is 33.6s .. 67.1s
+      const u8 max_latency_slot = 26;
+
+      // max number of listening ports we expect to see on the host
```
bobrik: nit: uppercase M.

PR #53, opened 2019-05-07
add dname decoder

xofyarg:
This decoder will convert DNS qname wire format into human readable
"." notation domain name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -655,6 +655,20 @@ An example to match `1` to `read` and `2` to `write`:
 String decoder transforms possibly null terminated strings coming
 from the kernel into string usable for prometheus metrics.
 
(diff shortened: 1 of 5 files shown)


bobrik:
There's one complaint from the linter, LGTM otherwise. What metrics are you gathering with this?

xofyarg:
I'm mainly getting the count for DNS query names, with labels focus only on type A/AAAA/other. Because the constant stack restriction in ebpf VM, decode doesn't work there.

One of the build failed with an HTTP 502 error this time, don't think the linter is unhappy though, maybe you could help me rerun it?

bobrik:
I restarted the build and it's happy now. Thanks!

PR #50, opened 2019-04-04
upgrade to libbcc 0.9.0

jasonk000:
Upgrades libbcc to 0.9.0, fixes #49

PR #54, opened 2019-06-12
Use libbcc 0.10.0

alexclear:


bobrik:
I think more places need an update, see: #50.

bobrik:
Merged in #57.

PR #73, opened 2020-04-01
Fix bug for Perf_event struct.

rexrock:
There is a code error in Perf_event struct, it causes only the HARDWARE event to work properly


bobrik:
Thanks!

PR #76, opened 2020-05-24
examples: add tcpsynbl.yaml for tracing the TCP SYN backlog

kenju:
Background:

There are two types of backlogs when TCP connection happens:

- SYN Backlog, which buffers connections on receiving SYN packets
- Listen Backlog, which buffers connections after sockets are ESTABLISHED but before accepted by applications

In the world of web programming, it is often userful to monitor the size of backlogs,
and find out whether we need to increase the number of backlogs limit.

ebpf_exporter achieves that you can collect those metrics in a histogram and export,
and can nicely visualize in Grafana dashboard for example.

---

@bobrik Hi 👋 I have been looking for collecting the TCP SYN backlog, and found out that `cloudflare/ebpf_exporter` let me to monitor the information, and can integrate with our alerting system smoothly using Prometheus and AlertManager 👍 

So I have just wondered if we can add a simple example for tracing the TCP SYN backlog in this repository. What do you think?

### DIFF
--- a/examples/tcpsynbl.yaml
+++ b/examples/tcpsynbl.yaml
@@ -1,0 +1,30 @@
+programs:
+  - name: tcpsynbl
+    metrics:
+      histograms:
+        - name: tcp_syn_backlog
+          help: trace the TCP SYN backlog size
+          table: bucket
+          bucket_type: linear
+          bucket_min: 0
+          bucket_max: 1024 # `sysctl net.ipv4.tcp_max_syn_backlog`
+          bucket_multiplier: 1
+          labels:
+            - name: bucket
+              size: 8
+              decoders:
+                - name: uint
+    kprobes:
+      tcp_v4_syn_recv_sock: do_count
+      tcp_v6_syn_recv_sock: do_count
+    code: |
+      #include <net/sock.h>
+
+      // Histograms to record TCP SYN backlog size
+      BPF_HISTOGRAM(bucket, u64);
+
+      int do_count(struct pt_regs *ctx, struct sock *sk) {
+          u64 backlog = bpf_log2l(sk->sk_ack_backlog);
+          bucket.increment(backlog);
+          return 0;
+      }
--- a/examples/tcpsynbl.yaml
+++ b/examples/tcpsynbl.yaml
@@ -3,7 +3,7 @@ programs:
     metrics:
       histograms:
         - name: tcp_syn_backlog
-          help: trace the TCP SYN backlog size
+          help: TCP SYN backlog size
           table: bucket
           bucket_type: linear
           bucket_min: 0
--- a/examples/tcpsynbl.yaml
+++ b/examples/tcpsynbl.yaml
@@ -7,8 +7,8 @@ programs:
           table: bucket
           bucket_type: linear
           bucket_min: 0
-          bucket_max: 1024 # `sysctl net.ipv4.tcp_max_syn_backlog`
-          bucket_multiplier: 1
+          bucket_max: 20
+          bucket_multiplier: 50
           labels:
             - name: bucket
               size: 8


examples/tcpsynbl.yaml
```diff
@@ -0,0 +1,30 @@
...
+      tcp_v4_syn_recv_sock: do_count
+      tcp_v6_syn_recv_sock: do_count
+    code: |
+      #include <net/sock.h>
+
+      // Histograms to record TCP SYN backlog size
+      BPF_HISTOGRAM(bucket, u64);
+
+      int do_count(struct pt_regs *ctx, struct sock *sk) {
+          u64 backlog = bpf_log2l(sk->sk_ack_backlog);
```
kenju, line 27: `sk_ack_backlog` shows the `current listen backlog`

https://www.kernel.org/doc/html/latest/networking/kapi.html?highlight=backlog

kenju: By the way, you can check [`sk_max_ack_backlog`](https://www.kernel.org/doc/html/latest/networking/kapi.html?highlight=backlog) at the same time and check whether we hit the limit or not.

examples/tcpsynbl.yaml
```diff
@@ -0,0 +1,30 @@
...
+          labels:
+            - name: bucket
+              size: 8
+              decoders:
+                - name: uint
+    kprobes:
+      tcp_v4_syn_recv_sock: do_count
+      tcp_v6_syn_recv_sock: do_count
+    code: |
+      #include <net/sock.h>
```
kenju, line 21: 📝 `<net/sock.h>` is required to use [`struct sock`](https://github.com/torvalds/linux/blob/caffb99b6929f41a69edbb5aef3a359bf45f3315/include/net/sock.h#L346-L522) and reference to the [`sk_ack_backlog`](https://github.com/torvalds/linux/blob/caffb99b6929f41a69edbb5aef3a359bf45f3315/include/net/sock.h#L478) value later.

examples/tcpsynbl.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - name: tcpsynbl
+    metrics:
+      histograms:
+        - name: tcp_syn_backlog
+          help: trace the TCP SYN backlog size
+          table: bucket
+          bucket_type: linear
+          bucket_min: 0
+          bucket_max: 1024 # `sysctl net.ipv4.tcp_max_syn_backlog`
+          bucket_multiplier: 1
```
bobrik, line 11: This produces 1024 time series, which is a bit excessive.

You can get a decent approximation of the backlog with:

```yaml
bucket_min: 0
bucket_max: 20
bucket_multiplier: 50
```

examples/tcpsynbl.yaml
```diff
@@ -0,0 +1,30 @@
+programs:
+  - name: tcpsynbl
+    metrics:
+      histograms:
+        - name: tcp_syn_backlog
+          help: trace the TCP SYN backlog size
```
bobrik, line 6:
```suggestion
          help: TCP SYN backlog size
```

examples/tcpsynbl.yaml
kenju: You're right. Thank you for pointing this out

✅ https://github.com/cloudflare/ebpf_exporter/pull/76/commits/dee0154db5fbc98542a0779c8352075601b89f31

examples/tcpsynbl.yaml
kenju: ✅ https://github.com/cloudflare/ebpf_exporter/pull/76/commits/aa59706bf7c58268f83dd7933a60bfecf0a09165

bobrik:
Thanks!

kenju:
Thank you for the review :)

PR #78, opened 2020-05-30
dockerfile: build cmd/ and add the binary to CMD instruction

kenju:
It would be great to build cmd/ and have it as a CMD instruction.

About CMD:

    https://docs.docker.com/engine/reference/builder/#cmd

Example:

    docker build -t ebpf_exporter .
    docker run ebpf_exporter

---

@bobrik Could you review this PR please?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.travis.yml
+LICENSE
+Makefile
+README.md
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,10 @@ RUN apt-get update && \
     apt-get install -y linux-headers-amd64 && \
     curl -sL https://github.com/cloudflare/ebpf_exporter/files/3890546/libbcc_0.11.0-2_amd64.deb.gz | gunzip > /tmp/libbcc.deb && \
     dpkg -i /tmp/libbcc.deb
-
 COPY ./ /go/ebpf_exporter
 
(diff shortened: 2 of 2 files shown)


bobrik:
Currently we only use docker to build binaries, the image is not supposed to be used for running. 

You need to bind-mount a whole bunch of things to make it work and without documentation it's just more misleading than helpful. I also have concerns about cross-distro compatibility between Debian inside of the container and whatever runs outside on the host.

kenju:
> You need to bind-mount a whole bunch of things to make it work and without documentation it's just more misleading than helpful.

Oh yes [I know what you mean](https://github.com/kenju/service-mesh-patterns/blob/fd596ea4889421a7c7efaf7a940a1ca103b7e851/prometheus-bpf-exporter/docker-compose.yml#L12-L14), as described in [the BCC repository's quickstart guide](https://github.com/iovisor/bcc/blob/master/QUICKSTART.md). We need to bind-mount modules, kernes source code, etc.

> Currently we only use docker to build binaries, the image is not supposed to be used for running.

I get it 👌 Still I think adding CMD will not have any disadvantage, and might be useful if someone wants to run bcc from docker image on linux by binding-mount. What do you think? 🤔

kenju:
Now I'm closing this issue for now because it's getting stale. It might not be a perfect solution, but I've pushed the forked version to the docker image registry and using it so my issue has been solved. Thank you for the review.

PR #83, opened 2020-10-12
Add machine check events tracing example

AlexZzz:
Hi!
This is an example of how EDAC errors might be collected using ebpf-exporter.

It is a stable way to collect such errors, since ```CONFIG_EDAC_LEGACY_SYSFS``` is not set.


bobrik:
Thanks!

PR #84, opened 2020-11-02
Add "allow_unknown" option to "static_map" decoder to not add "unknown:" prefix to keys that are not specified

matthewbentley:
If this option is true, static_map will not prefix unknown keys with 'unknown:'

The specific use-case is having a ebpf program that adds a "port" label for tcp connections, but "collapses" all ephemeral ports (32768 to 60999) into one value (60999), and then ebpf_exporter changes the label value from 60999 to "ephemeral".

### DIFF
--- a/README.md
+++ b/README.md
@@ -671,10 +671,24 @@ An example to match `1` to `read` and `2` to `write`:
 - name: operation
   decoders:
     - name:static_map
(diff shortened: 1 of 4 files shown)


bobrik:
Makes sense, thanks!

PR #93, opened 2021-02-07
mcevents: fix the size of the fields, add missed semicolons

noevhlev:
Size of the fields msg and label аre mixed up.
Truncate label size to 32((EDAC_MC_LABEL_LEN + 1)), increase MSG size to 64.
Add missed semicolons.

### DIFF
--- a/examples/mcevents.yaml
+++ b/examples/mcevents.yaml
@@ -34,7 +34,7 @@ programs:
               decoders:
                 - name: uint
             - name: label
-              size: 100
+              size: 32
               decoders:
                 - name: string
             - name: msg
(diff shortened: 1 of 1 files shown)


bobrik:
@AlexZzz do you mind taking a look?

AlexZzz:
32 bytes is really enough for the label field.
`msg` field's size is a driver-specific one. In most cases messages are quite short, so maybe it's better to keep it 32 bytes long?

noevhlev:
I checked again. Yes, you seem to be right. The original msg size is returned.

AlexZzz:
Thanks, looks good.

Issue #92, opened 2021-01-30
Build error


hsane2001:
The following command fails to build:  **go get -u -v github.com/cloudflare/ebpf_exporter/...**
github.com/cloudflare/ebpf_exporter (download)
github.com/cloudflare/ebpf_exporter/vendor/github.com/iovisor/gobpf/bcc
# github.com/cloudflare/ebpf_exporter/vendor/github.com/iovisor/gobpf/bcc
go/src/github.com/cloudflare/ebpf_exporter/vendor/github.com/iovisor/gobpf/bcc/module.go:261:33: not enough arguments in call to _C2func_bpf_attach_uprobe
        have (_Ctype_int, uint32, *_Ctype_char, *_Ctype_char, _Ctype_ulong, _Ctype_int)
        want (_Ctype_int, uint32, *_Ctype_char, *_Ctype_char, _Ctype_ulong, _Ctype_int, _Ctype_uint)

After forcing an additional param '0' at the end of the function allows the build to go through: "go/src/github.com/cloudflare/ebpf_exporter/vendor/github.com/iovisor/gobpf/bcc/module.go:261:33"

res, err := C.bpf_attach_uprobe(C.int(fd), attachType, evNameCS, binaryPathCS, (C.uint64_t)(addr), (C.pid_t)(pid)**, 0)**

I am unable to use uprobes, could this be the reason?

Thanks.

bobrik:
BCC doesn't have a stable API, which means that it breaks from time to time. It seems that you're using BCC 0.17.0 or newer, while `epbf_exporter` expects 0.16.0. The following commit from gonbpf is needed (same as your change): 

* https://github.com/iovisor/gobpf/commit/dcacba608f5153b453bfbe30a2a915e8e867d468

This has nothing to do with `ebpf_exporter`, it's API between gobpf and bcc.

Mic92:
You can sue this in the meantime: https://github.com/cloudflare/ebpf_exporter/pull/96

PR #96, opened 2021-03-02
upgrade bcc bindings

Mic92:
Motivation is that the current version no longer works with my kernel (5.4.99).
The dockerfile also needs a bump but I don't know how to update libbcc there.

Fixes https://github.com/cloudflare/ebpf_exporter/issues/92


bobrik:
Thanks, incorporated this into #97.

PR #102, opened 2021-05-11
Support BCC perf maps

zuzzas:
1. New configuration parameter for counters: `PerfMap`. You can reference it via BPF_PERF_OUTPUT and perf_submit in the C code.
2. An oom kill example which utilizes perf maps to push OOM events.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"time"
 )
 
 // Config defines exporter configuration
@@ -40,10 +41,12 @@ type Metrics struct {
 
 // Counter is a metric defining prometheus counter
 type Counter struct {
-	Name   string  `yaml:"name"`
-	Help   string  `yaml:"help"`
-	Table  string  `yaml:"table"`
-	Labels []Label `yaml:"labels"`
+	Name                 string        `yaml:"name"`
+	Help                 string        `yaml:"help"`
+	Table                string        `yaml:"table"`
+	PerfMap              string        `yaml:"perf_map"`
+	PerfMapFlushDuration time.Duration `yaml:"perf_map_flush_duration"`
+	Labels               []Label       `yaml:"labels"`
 }
 
 // Histogram is a metric defining prometheus histogram
--- a/examples/oomkill.yaml
+++ b/examples/oomkill.yaml
@@ -43,6 +43,6 @@ programs:
               return;
           }
 
-          data.cgroup_id = mcg->css.cgroup->kn->id.id;
+          data.cgroup_id = mcg->css.cgroup->kn->id;
           events.perf_submit(ctx, &data, sizeof(data));
       }
--- a/examples/oomkill.yaml
+++ b/examples/oomkill.yaml
@@ -1,0 +1,48 @@
+programs:
+  # See:
+  # * https://github.com/iovisor/bcc/blob/master/tools/oomkill.py
+  # * https://github.com/iovisor/bcc/blob/master/tools/oomkill_example.txt
+  - name: oomkill
+    metrics:
+      counters:
+        - name: oom_kills
+          help: Count global and cgroup level OOMs
+          perf_map: events
+          labels:
+            - name: cgroup_path
+              size: 8
+              decoders:
+                - name: uint
+                - name: cgroup
+            - name: global_oom
+              size: 1
+              decoders:
+                - name: uint
+    kprobes:
+      oom_kill_process: count_ooms
+    code: |
+      #include <uapi/linux/ptrace.h>
+      #include <linux/oom.h>
+      #include <linux/memcontrol.h>
+
+      // we'll use "BPF_PERF_OUTPUT" map type here to avoid unbound cardinality
+      BPF_PERF_OUTPUT(events);
+
+      struct data_t {
+          u64 cgroup_id;
+          u8 global_oom;
+      };
+
+      void count_ooms(struct pt_regs *ctx, struct oom_control *oc, const char *message) {
+          struct data_t data = {};
+
+          struct mem_cgroup *mcg = oc->memcg;
+          if (!mcg) {
+              data.global_oom = 1;
+              events.perf_submit(ctx, &data, sizeof(data));
+              return;
+          }
+
+          data.cgroup_id = mcg->css.cgroup->kn->id.id;
+          events.perf_submit(ctx, &data, sizeof(data));
+      }
--- a/exporter/perf_map.go
+++ b/exporter/perf_map.go
@@ -1,0 +1,120 @@
+package exporter
+
+import (
+	"log"
+	"time"
+
+	"github.com/cloudflare/ebpf_exporter/config"
+	"github.com/cloudflare/ebpf_exporter/decoder"
+	"github.com/iovisor/gobpf/bcc"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type PerfMapSink struct {
+	counterConfig config.Counter
+	counterVec    *prometheus.CounterVec
+	dropCounter   prometheus.Counter
+}
+
+func NewPerfMapSink(decoders *decoder.Set, module *bcc.Module, counterConfig config.Counter) *PerfMapSink {
+	var (
+		receiveCh = make(chan []byte)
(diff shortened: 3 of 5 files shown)


zuzzas:
Now that I think of it, there are other counting programs, such as [cgroup.yaml](https://github.com/cloudflare/ebpf_exporter/blob/master/examples/cgroup.yaml), that can benefit from it. This approach removes the unbounded growth of BPF_HASHes.

exporter/perf_map.go
```diff
@@ -0,0 +1,99 @@
...
+	sink := &PerfMapSink{
+		counterConfig: counterConfig,
+	}
+	sink.resetCounterVec()
+
+	table := bcc.NewTable(module.TableId(counterConfig.PerfMap), module)
+
+	perfMap, err := bcc.InitPerfMap(table, receiveCh, lostCh)
+	if err != nil {
+		log.Panicf("Can't init PerfMap: %s", err)
```
bobrik, line 33: Let's not panic and return error instead. We can do `log.Fatalf()` in the caller that cannot return errors.

exporter/perf_map.go
```diff
@@ -0,0 +1,99 @@
...
+				validDataSize += labelConfig.Size
+			}
+
+			labelValues, err := decoders.DecodeLabels(rawBytes[:validDataSize], counterConfig.Labels)
+			if err != nil {
+				if err == decoder.ErrSkipLabelSet {
+					continue
+				}
+
+				log.Printf("failed to decode labels: %s", err)
```
bobrik, line 51: Log messages should start with an uppercase letter, while errors go with lowercase.

exporter/perf_map.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+
+			sink.counterVec.WithLabelValues(labelValues...).Inc()
+
+		}
+	}(sink, counterConfig, receiveCh)
+
+	go func(sink *PerfMapSink) {
```
bobrik, line 59: Why do we need to reset counters every hour here? Should the time be configurable?

exporter/perf_map.go
```diff
@@ -0,0 +1,99 @@
...
+
+type PerfMapSink struct {
+	counterConfig config.Counter
+	counterVec    *prometheus.CounterVec
+}
+
+func NewPerfMapSink(decoders *decoder.Set, module *bcc.Module, counterConfig config.Counter) *PerfMapSink {
+	var (
+		receiveCh = make(chan []byte)
+		lostCh    = make(chan uint64)
```
bobrik, line 21: Do we need to drain `lostCh`?

examples/oomkill.yaml
```diff
@@ -0,0 +1,47 @@
...
+              decoders:
+              - name: uint
+    kprobes:
+      oom_kill_process: count_ooms
+    code: |
+      #include <uapi/linux/ptrace.h>
+      #include <linux/oom.h>
+      #include <linux/memcontrol.h>
+
+      BPF_PERF_OUTPUT(events);
```
bobrik, line 28: Could you add a note why `BPF_PERF_OUTPUT` is preferred here over a plain `BPF_HASH`? Is it unbound cardinality of the data or something else?

config/config.go
```diff
@@ -35,10 +35,11 @@ type Metrics struct {
...
 // Counter is a metric defining prometheus counter
 type Counter struct {
-	Name   string  `yaml:"name"`
-	Help   string  `yaml:"help"`
-	Table  string  `yaml:"table"`
-	Labels []Label `yaml:"labels"`
+	Name    string  `yaml:"name"`
+	Help    string  `yaml:"help"`
+	Table   string  `yaml:"table"`
+	PerfMap string  `yaml:"perf_map"`
```
bobrik, line 41: Let's document this in `README.md` as well. It's probably worth mentioning this in `counter` section and link to a new section describing the use case.

examples/oomkill.yaml
```diff
@@ -0,0 +1,47 @@
...
+    metrics:
+      counters:
+        - name: oom_kills
+          help: Count global and cgroup level OOMs
+          perf_map: events
+          labels:
+            - name: cgroup_path
+              size: 8
+              decoders:
+              - name: uint
```
bobrik, line 15: List items are indented with 2 spaces in other places, let's stick to that convention.

bobrik (changes requested):
Intention looks good (if I understand it correctly), but some changes and docs are needed.

decoder/cgroup.go
```diff
@@ -1,15 +1,15 @@
- package decoder
+package decoder
```
bobrik, line 1: Could you open a separate PR for this to keep things separate?

exporter/perf_map.go
zuzzas: Done.

exporter/perf_map.go
zuzzas: There is no error return up the call stack. Should I add them? I've changed this to `log.Fatalf` in the meantime.

exporter/perf_map.go
zuzzas: We do. I've added a Prometheus counter to count them.

examples/oomkill.yaml
zuzzas: Precisely. There could be thousands of cgroup_ids across hundreds of nodes. I've added Prometheus counter flushing for this exact reason.

examples/oomkill.yaml
zuzzas: Done.

exporter/perf_map.go
zuzzas: For the same reason as using a perf map: to avoid unbound label cardinality. I could switch this from `time.Duration` to counting currently active counters and drop them once they cross a threshold. Should be easier to forecast growth.

config/config.go
zuzzas: Done in a bit more concise way. Would that suffice?

decoder/cgroup.go
zuzzas: https://github.com/cloudflare/ebpf_exporter/pull/105

zuzzas: There are no changes to the cgroup decoder. It should work just the way it is.

examples/oomkill.yaml
```diff
@@ -0,0 +1,48 @@
...
+          struct data_t data = {};
+
+          struct mem_cgroup *mcg = oc->memcg;
+          if (!mcg) {
+              data.global_oom = 1;
+              events.perf_submit(ctx, &data, sizeof(data));
+              return;
+          }
+
+          data.cgroup_id = mcg->css.cgroup->kn->id.id;
```
bobrik, line 46: We should probably aim for more recent kernels in examples, and since v5.5 it's just `id`:

* https://github.com/torvalds/linux/commit/67c0496e87d193b8356d2af49ab95e8a1b954b3c

I tested on v5.10:

```
$ sudo systemd-run --scope -p MemoryMax=2K ls
Killed
$ sudo systemd-run --scope -p MemoryMax=2K ls
Killed
```

```
$ curl -s http://localhost:9435/metrics | grep ^ebpf
ebpf_exporter_ebpf_programs{function="count_ooms",program="oomkill",tag="6539c1cc556ca54a"} 1
ebpf_exporter_enabled_programs{name="oomkill"} 1
ebpf_exporter_oom_kills{cgroup_path="/sys/fs/cgroup/system.slice/run-r32c99c50a92f4d9bb1db9bde34d94637.scope",global_oom="0"} 1
ebpf_exporter_oom_kills{cgroup_path="/sys/fs/cgroup/system.slice/run-r98efd5fe68254f9c827b88ba0b0b875e.scope",global_oom="0"} 1
```

LGTM otherwise.

```suggestion
          data.cgroup_id = mcg->css.cgroup->kn->id;
```

zuzzas:
In addition to your latest comment, I think I'll add an OOM example with [tracepoints](https://lore.kernel.org/lkml/20170530185231.GA13412@castle/). I believe it could be useful to illustrate both approaches to this problem in the examples. That's why I'll do it in a follow-up PR.

bobrik (approved):

bobrik:
Thanks!

PR #104, opened 2021-06-23
Run some checks on parsed configs

ZekeMedley:
This adds two simple sanity checks that run on configs before they get compiled.


bobrik:
I added another commit to push config validation into `exporter.New()` to make sure we don't create exporters with invalid configuration. I also dropped "sanity" from the commit message ([see here](https://developers.google.com/style/inclusive-documentation)).

Other than that looks good, thanks!

PR #106, opened 2021-09-09
Scan whole /sys/fs/cgroup directory in the cgroup decoder and use a faster filepath.WalkDir function

zuzzas:
1. Scan whole /sys/fs/cgroup directory in the cgroup decoder.
  This resolves a problem when you are using a Hybrid systemd cgroup mode.
  Reference: https://systemd.io/CGROUP_DELEGATION/
2. Use a newer `filepath.WalkDir` function to traverse the cgroup hierarchy. It provides a major speedup to the refresh process.

### DIFF
--- a/decoder/cgroup.go
+++ b/decoder/cgroup.go
@@ -2,8 +2,8 @@ package decoder
 
 import (
 	"fmt"
+	"io/fs"
 	"log"
-	"os"
 	"path/filepath"
 	"strconv"
 
@@ -48,7 +48,7 @@ func (c *CGroup) refreshCache() error {
 
 	cgroupPath := "/sys/fs/cgroup"
 
-	return filepath.Walk(cgroupPath, func(path string, info os.FileInfo, err error) error {
+	return filepath.WalkDir(cgroupPath, func(path string, info fs.DirEntry, err error) error {
 		if err != nil {
 			return err
 		}
--- a/decoder/cgroup.go
+++ b/decoder/cgroup.go
@@ -18,7 +18,7 @@ type CGroup struct {
 }
 
 // Decode transforms cgroup id to path in cgroupfs
-func (c *CGroup) Decode(in []byte, conf config.Decoder) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


zuzzas:
@bobrik

bwplotka:
I think this will only work if users use cgroups v1. For v2 cgroups we need to look in `unified` no?

bwplotka:
Context: I have issues with latency of this endpoint - it takes a huge amount of time to traverse through that `refresh` when new `cgroup id` appears. We need faster mechanism here.

bwplotka:
Ideally with this: https://github.com/iovisor/bpftrace/issues/1500

bwplotka:
Added issue: https://github.com/cloudflare/ebpf_exporter/issues/109

zuzzas:
@bwplotka 
Thanks for the comment. There is a faster implementation of Walk if you are not concerned with `Stat()`ing every single file. I see a major speedup on a Linux laptop with the newest commit.

```
root@b5dace1effc1:/src/decoder# go test -run=XXX -bench=. .
goos: linux
goarch: amd64
pkg: github.com/cloudflare/ebpf_exporter/decoder
cpu: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
BenchmarkCgroupRefresh-8              55          19956497 ns/op
PASS
ok      github.com/cloudflare/ebpf_exporter/decoder     1.142s
root@b5dace1effc1:/src/decoder# go test -run=XXX -bench=. .
goos: linux
goarch: amd64
pkg: github.com/cloudflare/ebpf_exporter/decoder
cpu: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
BenchmarkCgroupRefresh-8              24          47062521 ns/op
PASS
ok      github.com/cloudflare/ebpf_exporter/decoder     1.838s
```

bobrik:
LGTM. @bwplotka, are you happy with this?

bobrik (approved):

PR #126, opened 2021-11-04
Support to add const labels for metrics

wenlxie:
This is to support user cam add some const labels for each metrics.
eg: user can add the node fqdn info when running in k8s world.

### DIFF
--- a/cmd/ebpf_exporter/main.go
+++ b/cmd/ebpf_exporter/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"log"
 	"net/http"
+	"strings"
 
 	"github.com/cloudflare/ebpf_exporter/config"
 	"github.com/cloudflare/ebpf_exporter/exporter"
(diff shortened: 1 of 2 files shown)


bobrik:
This sort of thing is usually accomplished from the scraping side:

* https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs

Do you have an example of any other exporter in prometheus org that allows const labels instead?

wenlxie:
@bobrik Thanks for the information.

Maybe this can be the example: https://github.com/prometheus/influxdb_exporter/blob/master/main.go#L275 which consider about the add const labels to the metrics. 

For the cases that I mentioned, `metric_relabel_configs` can be the solution.  Actually this is also the way we are using now. With the const labels which can be specified by the args, which can bring more flexiable for users.

timbuchwaldt:
Because I needed exactly this on my exporter and was about to upvote:
```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: ebpf-prometheus
spec:
  selector:
    matchLabels:
      app: ebpf-exporter
  namespaceSelector:
    matchNames:
    - ebpf-exporter
  podMetricsEndpoints:
  - port: prometheus
    interval: 10s
    relabelings:
    - action: replace
      sourceLabels:
      - __meta_kubernetes_pod_node_name
      targetLabel: instance    
```

PR #130, opened 2021-12-28
Make ebpf exporter to support CORE

wenlxie:
This PR is for ebpf exporter to support CORE. 
This can make ebpf exporter to get CORE benefits and tolerate kernel changes.

I'd believe that there are more things need to do with this so please share your suggestions. Thanks.


wenlxie:
@bobrik Can you help to take a look?

bobrik:
There are lots of changes here and many of them are unrelated:

* Removed comments that still apply (starting with `Dockerfile`)
* Inconsistent indentation in C files
* Either missing or multiple newlines at the end of files
* Broken Go formatting, even on the lines you didn't really change
* Error strings starting with capital letters (only log lines should)
* Functions like `tableValues` moved around and generating large diff with no logical changes

Could you take care of this first to make it easier to look at the actual changes?

There are SPDX headers for LGPL, I'll need to ask legal people whether we can accept it.

I would also like to see some documentation in the README on how to use this.

wenlxie:
@bobrik Thanks for the suggestion. I will take care of them. 
After move to CORE mode,  BCC related will be removed,  so I raised this PR for your to review first. Move from BCC to CORE is a right direction, but need your agreement with this.  
Thanks again.

wenlxie:
How to run ebpf_exporter with CORE mode?
1. Run make to have generate the image directly.
2. Run the image with with configuration like:
```
        - --web.listen-address=:10678
        - --config.file=/examples/CORE/config.yaml
 ```

File: Dockerfile [hunk 1]
```diff
@@ -1,9 +1,6 @@
-# Building on Ubuntu Bionic (18.04) and supporting glibc 2.27. This allows
```
bobrik, line 1: Does this comment no longer apply?

Dockerfile
```diff
@@ -1,9 +1,6 @@
...
-# the following distros (and newer versions) to run the resulting binaries:
-# * Ubuntu Bionic (2.27)
-# * Debian Buster (2.28)
-# * CentOS 8 (2.28)
-FROM ubuntu:bionic as builder
+FROM ubuntu:20.04 as builder
```
bobrik, line 1: Any reason to change this?

Dockerfile
```diff
@@ -1,9 +1,6 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
```
bobrik, line 3: It worked without this env var, do we need it for something specific with your changes?

Dockerfile
```diff
@@ -15,29 +12,55 @@ RUN cd /root/bcc && \
...
 
-FROM ubuntu:bionic
+RUN cd /root && git clone https://github.com/libbpf/libbpf.git && \
+    cd /root/libbpf && git checkout 030ff87857090ae5c9d74859042d05bfb3b613a2  && cd src && mkdir build root && BUILD_STATIC_ONLY=y OBJDIR=build DESTDIR=root make install
+
+FROM ubuntu:20.04 as compiler
+ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends git build-essential libelf1 software-properties-common
```
bobrik, line 21: Please make sure that Dockerfile is still operational:

```
Step 18/32 : RUN cd /go/ebpf_exporter && CC=gcc CGO_CFLAGS="-I /usr/include/bpf" CGO_LDFLAGS="-lbpf" GOPATH="" GOPROXY="off" GOFLAGS="-mod=vendor" go install -v -ldflags="     -X github.com/prometheus/common/version.Version=$(git describe)     -X github.com/prometheus/common/version.Branch=$(git rev-parse --abbrev-ref HEAD)     -X github.com/prometheus/common/version.Revision=$(git rev-parse --short HEAD)     -X github.com/prometheus/common/version.BuildUser=docker@$(hostname)     -X github.com/prometheus/common/version.BuildDate=$(date --iso-8601=seconds)     " ./cmd/ebpf_exporter
 ---> Running in 7f291d3a0ba1
/bin/sh: 1: git: not found
/bin/sh: 1: git: not found
/bin/sh: 1: git: not found
```

Dockerfile
```diff
@@ -15,29 +12,55 @@ RUN cd /root/bcc && \
...
+
+COPY --from=builder /root/bcc/libbcc_*.deb /tmp/libbcc.deb
+COPY --from=builder /root/bcc/libbpf-tools  /tmp/libbpf-tools
+COPY --from=builder /root/libbpf/ /tmp/libbpf
+COPY --from=compiler /root/go/bin/ebpf_exporter /ebpf_exporter
+COPY --from=compiler /go/ebpf_exporter/examples /examples
+
+RUN cd /tmp/libbpf/src && make install
+RUN dpkg -i /tmp/libbcc.deb
+RUN find /tmp -name "*.git*"|xargs rm -Rf
```
bobrik, line 63: Why remove stuff here?

Dockerfile
```diff
@@ -15,29 +12,55 @@ RUN cd /root/bcc && \
...
+RUN find /examples -name "*.git*"|xargs rm -Rf
 
-RUN /root/go/bin/ebpf_exporter --version
+ENV LD_LIBRARY_PATH=/lib64
```
bobrik, line 66: The docker image is not supposed to be used directly, but it's used to build the binaries. The `RUN` command was there to make sure that the binaries were operational.

I'm not sure what's the purpose of `LD_LIBRARY_PATH`.

Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: vendor
 vendor:
 	docker build -t ebpf-exporter-build .
-	docker run --rm -v $(CURDIR):/go/ebpf_exporter --workdir /go/ebpf_exporter --entrypoint /bin/bash ebpf-exporter-build -c "go get -u ./... && go mod vendor && go mod tidy"
+	docker run --rm -v $(CURDIR):/go/ebpf_exporter --workdir /go/ebpf_exporter --entrypoint /bin/bash ebpf-exporter-build -c "CGO_CFLAGS="-I /usr/include/bpf" CGO_LDFLAGS="-lbpf" go get -u ./... && go mod vendor && go mod tidy"
```
bobrik, line 4: Any non-standard env variables should be explained. If they are sprinkled in multiple places and required for the build, then it's probably best to document the reasons in the README. I was able to find the following issue for `LDFLAGS`:

* https://github.com/aquasecurity/libbpfgo/issues/1

Not really sure why `CFLAGS` is necesssary.

Dockerfile
```diff
@@ -15,29 +12,55 @@ RUN cd /root/bcc && \
...
+COPY --from=builder /root/libbpf/ /tmp/libbpf
 
+RUN cd /tmp/libbpf/src && make install
 RUN dpkg -i /tmp/libbcc.deb
 
 COPY ./ /go/ebpf_exporter
 
-RUN cd /go/ebpf_exporter && \
-    GOPROXY="off" GOFLAGS="-mod=vendor" go install -v -ldflags=" \
+RUN cd /go/ebpf_exporter && CC=gcc CGO_CFLAGS="-I /usr/include/bpf" CGO_LDFLAGS="-lbpf" GOPATH="" GOPROXY="off" GOFLAGS="-mod=vendor" go install -v -ldflags=" \
```
bobrik, line 38: Why do you need to override `CC`?

examples/CORE/Makefile
```diff
@@ -0,0 +1,28 @@
+# SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+OUTPUT := .output
+CLANG ?= clang
+LLVM_STRIP ?= llvm-strip
+BPFTOOL ?= bin/bpftool
```
bobrik, line 5: This is not used anywhere

examples/CORE/Makefile
```diff
@@ -0,0 +1,28 @@
...
+LIBBPF_OBJ := /tmp/libbpf-tools/libbpf.a
```
bobrik, line 6: Also unused

examples/CORE/Makefile
```diff
@@ -0,0 +1,28 @@
...
+CFLAGS := -g -O2 -Wall
+ARCH := $(shell uname -m | sed 's/x86_64/x86/')
```
bobrik, line 8: This needs a comment explaining the reason.

examples/CORE/Makefile
```diff
@@ -0,0 +1,28 @@
...
+INCLUDES := -I"/tmp/libbpf-tools/.output" -I "/tmp/libbpf-tools/$(ARCH)"
```
bobrik, line 9: What are these includes in `/tmp`?

examples/CORE/Makefile
```diff
@@ -0,0 +1,28 @@
+# SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause)
+OUTPUT := .output
+CLANG ?= clang
+LLVM_STRIP ?= llvm-strip
+BPFTOOL ?= bin/bpftool
+LIBBPF_OBJ := /tmp/libbpf-tools/libbpf.a
+CFLAGS := -g -O2 -Wall
```
bobrik, line 7: These seem reasonable, but you don't pass different ones below.

examples/CORE/biolatency.h
```diff
@@ -0,0 +1,87 @@
...
+	[REQ_OP_DISCARD] = "Discard",
+	[REQ_OP_SECURE_ERASE] = "SecureErase",
+	[REQ_OP_ZONE_RESET] = "ZoneReset",
+	[REQ_OP_WRITE_SAME] = "WriteSame",
+	[REQ_OP_ZONE_RESET_ALL] = "ZoneResetAll",
+	[REQ_OP_WRITE_ZEROES] = "WriteZeroes",
+	[REQ_OP_ZONE_OPEN] = "ZoneOpen",
+	[REQ_OP_ZONE_CLOSE] = "ZoneClose",
+	[REQ_OP_ZONE_FINISH] = "ZoneFinish",
+	[REQ_OP_SCSI_IN] = "SCSIIn",
```
bobrik, line 81: You might want to `#ifdef REQ_OP_SCSI_IN` the next two lines:

* https://github.com/torvalds/linux/commit/da6269da4cfe29f484e8fd27c1496b81b47e2499

examples/CORE/biolatency.bpf.c
```diff
@@ -0,0 +1,175 @@
+/* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
+#include <vmlinux.h>
```
bobrik, line 2: Where is this file supposed to come from?

examples/CORE/biolatency.yaml
```diff
@@ -0,0 +1,123 @@
...
+                    32: SCSIIn
+                    33: SCSIOut
+                    34: DrvIn
+                    35: DrvOut
+            - name: bucket
+              size: 4
+              decoders:
+                - name: uint
+    raw_tracepoints:
+      inet_csk_accept: rawtracepoint__block_rq_issue
```
bobrik, line 121: This doesn't seem like the right tracepoint name.

examples/CORE/config.yaml
```diff
@@ -0,0 +1,210 @@
+programs:
```
bobrik, line 1: This file looks like a combination of multiple files, should it be here?

examples/CORE/mm.yaml
```diff
@@ -0,0 +1,49 @@
...
+                - name: uint
+                - name: static_map
+                  static_map:
+                    0: DMA
+                    1: DMA32
+                    2: NORMAL
+                    3: MOVEABLE
+                    4: DEVICE
+    tracepoints:
+      "compaction:mm_compaction_kcompactd_wake" tracepoint__compaction__mm_compaction_kcompactd_wake
```
bobrik, line 49: Looks like there's a missing `:` here. If you don't need the quotes around the keys, I'd recommend to drop them.

examples/CORE/shrinklat.yaml
```diff
@@ -0,0 +1,23 @@
...
+          labels:
+            - name: bucket
+              size: 8
+              decoders:
+                - name: uint
+    kprobes:
+      shrink_node: kprobe__shrink_node
+    kretprobes:
+      shrink_node: kretprobe__shrink_node
+
```
bobrik, line 23: Trailing newline

exporter/exporter.go
```diff
@@ -48,62 +52,63 @@ func New(cfg config.Config) (*Exporter, error) {
...
 		decoders:            decoder.NewSet(),
 	}, nil
 }
 
-// Attach injects eBPF into kernel and attaches necessary kprobes
-func (e *Exporter) Attach() error {
+// Support CORE Mode
+func (e *Exporter) Attach(configPath string) error {
 	for _, program := range e.config.Programs {
-		if _, ok := e.modules[program.Name]; ok {
```
bobrik, line 70: Why remove this check?

exporter/exporter.go
```diff
@@ -48,62 +52,63 @@ func New(cfg config.Config) (*Exporter, error) {
...
-			return fmt.Errorf("multiple programs with name %q", program.Name)
-		}
+		bpfObjPath := program.Name
+		fmt.Printf("%v", bpfObjPath)
```
bobrik, line 75: This seems like a debug leftover

exporter/exporter.go
```diff
@@ -226,15 +228,12 @@ func (e *Exporter) collectCounters(ch chan<- prometheus.Metric) {
 			if len(counter.PerfMap) != 0 {
 				continue
 			}
-
 			tableValues, err := e.tableValues(e.modules[program.Name], counter.Table, counter.Labels)
 			if err != nil {
 				log.Printf("Error getting table %q values for metric %q of program %q: %s", counter.Table, counter.Name, program.Name, err)
 				continue
 			}
-
```
bobrik, line 235: Please don't remove newlines like that.

exporter/exporter.go
```diff
@@ -21,12 +25,12 @@ const prometheusNamespace = "ebpf_exporter"
 // Exporter is a ebpf_exporter instance implementing prometheus.Collector
 type Exporter struct {
 	config              config.Config
-	modules             map[string]*bcc.Module
+	modules             map[string]*bpf.Module
 	perfMapCollectors   []*PerfMapSink
 	kaddrs              map[string]uint64
 	enabledProgramsDesc *prometheus.Desc
 	programInfoDesc     *prometheus.Desc
-	programTags         map[string]map[string]uint64
```
bobrik, line 29: Why remove tags?

bobrik (changes requested):
Unfortunately, I was not able to build this either with the Dockerfile or in a Debian Bullseye VM with libbpf from backports. Happy to have another look when I can build and run it myself.

Dockerfile
wenlxie: Yes, /usr/lib/pbuilder/pbuilder-satisfydepends will install the package tzdata, and tzdata will request to set the Geographic area and time zonem, so this env is needed.

Dockerfile
wenlxie: `git` info is treated as vulnerability in some companies's image security policy, so remove this.

Dockerfile
wenlxie: There will be errors like when use lower LLVM version:
```
bindsnoop.bpf.c:76:20: error: use of unknown builtin '__builtin_preserve_access_index' [-Wimplicit-function-declaration]
.output/bpf/bpf_core_read.h:404:2: note: expanded from macro 'BPF_CORE_READ'
        BPF_CORE_READ_INTO(&__r, (src), a, ##__VA_ARGS__);                  \
```

Dockerfile
wenlxie: ebpf_exporter is built by container, so I'd think ebpf_exporter be better to run in a container,  it is no need to care about the os/libc version on the host.

Dockerfile
wenlxie: Fixed

exporter/exporter.go
wenlxie: libbpfgo doesn't have interface to get the tags for program. 
so I changed this to expose the functions

examples/CORE/biolatency.bpf.c
wenlxie: https://github.com/iovisor/bcc/tree/master/libbpf-tools/x86

examples/CORE/config.yaml
wenlxie: It is better to have a config file that can be used by user directly, so I create this file.

Dockerfile
wenlxie:
```
./ebpf_exporter: error while loading shared libraries: libbpf.so.0: cannot open shared object file: No such file or directory```
```
ls /lib64/
ld-linux-x86-64.so.2  libbpf.a  libbpf.so  libbpf.so.0  libbpf.so.0.5.0  pkgconfig
```
The binary can't find libbpf.so.0 if this env not set

File: examples/CORE/mm.yaml (hunk 18 above)
Comment (wenlxie) [reply examples/CORE/mm.yaml]: 1. Add the `:` 
2.  Remove the quotes

**Comment by wenlxie** [comment]
@bobrik  I addressed some comments today but not finished all of these, will continue to do this tomorrow.
Thanks .

File: Dockerfile (hunk 1 above)
Comment (bobrik) [reply Dockerfile:1]: Everything is built in a container these days, because it's convenient and somewhat reproducible. We only distribute a binary, though, not a docker image. I would like to keep the binary with max compatibility, as people seem to depend on it: #115.

For people who don't run Docker, it's certainly not an option to install it just for ebpf_exporter.

File: Dockerfile (hunk 2 above)
Comment (bobrik) [reply Dockerfile:1]: Let's use https://apt.llvm.org/bionic/ if Ubuntu repos don't have what we need.

File: Dockerfile (hunk 3 above)
Comment (bobrik) [reply Dockerfile:3]: `pbuilder-satisfydepends` was there before and it worked. Is it related to bionic -> focal bump then?

File: Dockerfile (hunk 5 above)
Comment (bobrik) [reply Dockerfile:63]: For people that do this, there's always an option to build from source. We don't need to do any cleanup in this image, as it's not intended as a distributable artifact. There are only two things of interest here:

* The image should build a binary
* The binary should be compatible with older libc (as mentioned in the very first comment)

File: Dockerfile (hunk 6 above)
Comment (bobrik) [reply Dockerfile:66]: Is there a way to statically link libbpf? I know that it's not really possible with libbcc, but it would be really nice if the produced binary worked without extra deps.

File: Dockerfile [hunk 24]
```diff
@@ -15,29 +12,56 @@ RUN cd /root/bcc && \
...
+COPY --from=compiler /go/ebpf_exporter/examples /examples
+
+RUN cd /tmp/libbpf/src && make install
+RUN dpkg -i /tmp/libbcc.deb
+RUN find /tmp -name "*.git*"|xargs rm -Rf
+RUN find /examples -name "*.git*"|xargs rm -Rf
 
-RUN /root/go/bin/ebpf_exporter --version
+ENV LD_LIBRARY_PATH=/lib64
+CMD /ebpf_exporter  --web.listen-address=:10678  --config.file=/examples/CORE/config.yaml
```
bobrik, line 67: As mentioned above, this image is not intended for direct consumption, so we don't need this line.

There's also a standard port allocated for ebpf_exporter:

* https://github.com/prometheus/prometheus/wiki/Default-port-allocations

Makefile
```diff
@@ -1,9 +1,18 @@
...
+endif
+
+IMAGE = ebpf-exporter:$(VERSION)
+
+container:
+	docker build -t $(IMAGE) .
+.PHONY: container
+
+push: container
+	docker push $(IMAGE)
```
bobrik, line 12: I do not intend to push the image anywhere, so let's not add things that will not be used.

examples/CORE/biolatency.bpf.c
bobrik: What should an end user do when all they have is a binary or the source of this repo?

examples/CORE/config.yaml
bobrik: All examples are intended to be isolated single programs per config. The end users can combine multiple if they choose so, but that's something that happens outside of this repo.

exporter/exporter.go
bobrik: Could you open an issue for libbpfgo to add support for that? IIRC, you can only see program tags in `perf` output, so having tags to match those back to ebpf_exporter programs is useful to have.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/alecthomas/units v0.0.0-20210912230133-d1bdfacee922 // indirect
+	github.com/aquasecurity/libbpfgo v0.2.1-libbpf-0.4.0
```
bobrik, line 7: Please run `go mod tidy`, there's a trailing space in this line.

Dockerfile
bobrik: I'm getting a linker error now:

```
# runtime/cgo
/usr/bin/ld: cannot find -lbpf
collect2: error: ld returned 1 exit status
```

Not sure if the fact that I'm on aarch64 has anything to do with it.

exporter/exporter.go
wenlxie: IC, I didn't realized that tag can be used for this.  Let me make libbpfgo  to support tag first.

rob-scheepens:
Due to not wanting to install Docker on all machines I intend to run ebpf_exporter on I tried to build on Bullseye vm as well. When running "./scripts/build-deb.sh release" I got:
```
dpkg-checkbuilddeps: error: Unmet build dependencies: python-netaddr

When trying to install python-netaddr:
```
root@rob-bullseye:~/bcc# apt-get install -y python-netaddr
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package python-netaddr is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python3-netaddr

E: Package 'python-netaddr' has no installation candidate

Installing python3-netaddr works fine on Bullseye, but that doesn't address the dependency issue.

bobrik:
@rob-scheepens, I think bcc build script checks out `HEAD`, so you need to commit your change to apply it.

Not sure how it's related to CO-RE.

rob-scheepens:
@bobrik : Using the Dockerfile from https://github.com/wenlxie/ebpf_exporter/blob/upstream.master.core/Dockerfile I attempted to convert this to commands that can be scripted and run on a Debian Bullseye vm. All commands for now assume the user is root.
```
export PATH="/usr/lib/go-1.17/bin:$PATH"
export DEBIAN_FRONTEND=noninteractive
export LD_LIBRARY_PATH=/lib64

echo 'deb http://deb.debian.org/debian bullseye-backports main contrib non-free
deb-src http://deb.debian.org/debian bullseye-backports main contrib non-free' >> /etc/apt/sources.list.d/backports.list

sed -i "s/bullseye main/bullseye main non-free/" /etc/apt/sources.list

apt-get update

apt-get --no-install-recommends install -y sudo devscripts build-essential fakeroot pbuilder aptitude git openssh-client ca-certificates git pkg-config libelf-dev libz-dev software-properties-common clang gnupg python3-netaddr pkg-config libz-dev software-properties-common bpftool python2 python3 dh-python libclang-9-dev libclang-common-9-dev libclang-cpp9 libclang1-9 libllvm9 llvm-9 llvm-9-dev llvm-9-runtime llvm-9-tools netperf python3-pyroute2 clang-format bison flex libfl-dev libedit-dev arping iperf ethtool python3 python3-doc python3-tk python3-venv python3.9-venv python3.9-doc cmake luajit libluajit-5.1-dev debhelper
apt-get install -y -t bullseye-backports golang-1.17 libbpfcc-dev libbpfcc bpfcc-tools

git clone --branch=v0.22.0 --depth=1 https://github.com/iovisor/bcc.git /root/bcc && \
git -C /root/bcc submodule update --init --recursive

git clone https://github.com/libbpf/libbpf.git /root/libbpf && \
cd /root/libbpf && git checkout 030ff87857090ae5c9d74859042d05bfb3b613a2 && cd src && mkdir build root && BUILD_STATIC_ONLY=y OBJDIR=build DESTDIR=root make install
At this point a modification to the debian/control file is needed, see https://github.com/rob-scheepens/bcc/commit/6335995dec421e60298651cb6120d5bec247bd23 and my previous commit on that file, changing python2 dependencies as these do not work on Bullseye anymore. Script to make the changes and commit them (since the build-deb.sh script checks out HEAD):
cd /root/bcc/
git config --global user.email "user@[email.com](http://email.com/)"
git config --global [user.name](http://user.name/) "nutanix"
sed -i "s/python (>= 2.7)/python3/" debian/control
sed -i "s/python-netaddr/python3-netaddr/" debian/control
git add debian/control
git commit -m "python3 changes."
./scripts/build-deb.sh release

git clone https://github.com/wenlxie/ebpf_exporter.git /root/ebpf_exporter
cd /root/ebpf_exporter
git checkout upstream.master.core
```
This gets me up to the point where I run into the issue described in https://github.com/aquasecurity/libbpfgo/issues/1, where "go install" fails. I am trying to work out what to do about that.

wenlxie:
I update the code:
1. Remove bcc dependency
2. Add tag support back
3. Make lib to be static linked
4. benchmark folder removed because of 1, will add it back 

@bobrik  @rob-scheepens 
Now the ebpf_exporter is static compiled,  it works well in my env.

rob-scheepens:
@wenlxie : using your latest upstream.master.core I checked again and the script above still works.

When you say "Remove bcc dependency", does that mean that there is no need to have bcc installed/built anymore, and that I can remove it from the script? When I tested that, things failed.

wenlxie:
> When you say "Remove bcc dependency", does that mean that there is no need to have bcc installed/built anymore, and that I can remove it from the script? When I tested that, things failed.

It means remove the bcc codes from the code/vendor

But it still need the header file in bcc to compile the bpf binary.

rob-scheepens:
> Unfortunately, I was not able to build this either with the Dockerfile or in a Debian Bullseye VM with libbpf from backports. Happy to have another look when I can build and run it myself.

@bobrik : can you give it another try using the script I provided for Bullseye?

cc: @wenlxie

Dockerfile
rob-scheepens: @bobrik : could you give it a try with the script in https://github.com/cloudflare/ebpf_exporter/pull/130#issuecomment-1046722298 ?

bobrik:
I was able to build this locally with some modifications. My plan is to clean this up and see how we can get this merged.

Apologies for the delay.

wenlxie:
> I was able to build this locally with some modifications. My plan is to clean this up and see how we can get this merged.
> 
> Apologies for the delay.

@bobrik Thanks for the test,  please tell me with anything that I can help.

bobrik (approved):

bobrik:
https://github.com/aquasecurity/libbpfgo/pull/258 looks like it could be a good replacement for kaddr (or any argument) passing.

Not sure if we should wait for it though.

bobrik:
Many thanks @wenlxie! I'll try to roll it out internally to iron any kinks before tagging a proper v2.0.0 release.

bobrik:
It's out now: https://github.com/cloudflare/ebpf_exporter/releases/tag/v2.0.0

PR #135, opened 2022-03-22
examples: summarize TCP send/recv throughput by host

lubinszARM:
Add an example similar to bcc/tcptop, and add the function of getting
the container name.

Signed-off-by: Robin Lu <lubinsz@gmail.com>
Signed-off-by: Robin Lu <robin.lubin@shopee.com>

### DIFF
--- a/examples/container-tcptop.yaml
+++ b/examples/container-tcptop.yaml
@@ -1,0 +1,166 @@
+programs:
+  - name: container-tcptop
+    metrics:
+      counters:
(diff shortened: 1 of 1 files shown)


lubinszARM:
@bobrik 
Any comments?

lubinszARM:
cc @dqminh

bobrik:
It's not something I would use as is due to cardinality, but I think it's good to have as a starting point example. Thanks!

PR #136, opened 2022-04-18
Kernel version constraints and enabling programs individually

zuzzas:
1. Enable programs invidually. Do not fail whole process if one program fails validation/build. Reflect whether program exists or not in the `ebpf_exporter_enabled_programs` metric.
2. Allows specifying kernel version constraints. So that programs won't run on unexpected kernel versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,4 +128,4 @@ jobs:
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.42.1
+          version: v1.45.2
--- a/cmd/ebpf_exporter/main.go
+++ b/cmd/ebpf_exporter/main.go
@@ -3,14 +3,16 @@ package main
 import (
 	"log"
 	"net/http"
+	"regexp"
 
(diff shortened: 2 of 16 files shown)


zuzzas:
@bobrik 
I'll fix tests and write the docs. First I wanted to check whether you like the idea or not?

zuzzas:
@bobrik 
We've been using this patch since May on 200+ Kubernetes clusters. This allows us to write different eBPF programs for different kernel versions. Really useful for dynamic kprobe instrumentation.

Are you interested in accepting this PR and idea as a whole?

bobrik:
My plan is to land #130 first. I think it should unlock kernel constraints for you as well.

Not failing on some programs makes sense. I would postpone it until CO-RE to avoid conflicts.

bobrik:
See #143 for my kernel version constraint thoughts.

Having graceful failure to attach program still sounds good, so feel free to resolve conflicts for that.

bobrik:
I opened #161 that should address graceful handling of failure to attach. Let me know what you think.

zuzzas:
Nice. And I agree with your extensive point about kernel constraints. We'll manage the config externally.

PR #145, opened 2022-10-20
Use generic program attach mechanism

bobrik:
With `SEC()` macro present in bpf C code we no longer need a redundant mapping from kprobes/tracepoints/whatever to corresponding elf symbols.

We can just let libbpf attach as described.


bobrik:
Unfortunately, `SEC("perf_event")` isn't customizable enough out of the box.

We can try using `libbpf_register_prog_handler`:

* https://libbpf.readthedocs.io/en/latest/api.html#libbpf_8h_1a4e9b922125a34bb938514ebe29ec6a4d

Something like `SEC("perf_event/type=0x0,name=0x3,frequency=99")` for this existing config:

```yaml
      - type: 0x0 # HARDWARE
        name: 0x3 # PERF_COUNT_HW_CACHE_MISSES
        target: on_cache_miss
        sample_frequency: 99
```

bobrik:
See #154 for the custom `perf_event` handler.

PR #151, opened 2022-10-20
Drop inline vendor

SuperQ:
The Go community has settled on not using inline vendoring in repos now that we have Go modules with full direct and indirect dependencies tracked.

Signed-off-by: SuperQ <superq@gmail.com>


Makefile
```diff
@@ -1,5 +1,5 @@
 export CGO_LDFLAGS := -l bpf
-export GOFLAGS := -mod=vendor
+export GOFLAGS :=
```
bobrik, line 2: No need to keep it around, let's remove it altogether.

SuperQ: Sure, that seems reasonable. Sorry I didn't get back to this, I was traveling.

PR #158, opened 2022-10-23
Update Docker build

SuperQ:
Some simplification of the Dockerfile.
* Use official Go build image.
* Consolidate RUN steps to reduce layers.
* Use "distroless" for final image.

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,20 @@
-FROM ubuntu:22.04 as builder
+FROM golang:1.19-bullseye as builder
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends curl git ca-certificates build-essential libelf-dev gnupg2
+    apt-get install -y --no-install-recommends libelf-dev
 
-RUN echo 'deb https://ppa.launchpadcontent.net/longsleep/golang-backports/ubuntu jammy main' > /etc/apt/sources.list.d/golang-backports.list && \
-    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 52b59b1571a79dbc054901c0f6bc817356a3d45e && \
-    apt-get update && \
-    apt-get install -y --no-install-recommends golang-1.19-go
-
-RUN mkdir /build
-
-RUN git clone --branch v1.0.1 --depth 1 https://github.com/libbpf/libbpf.git /build/libbpf && \
-    make -C /build/libbpf/src BUILD_STATIC_ONLY=y LIBSUBDIR=lib install
-
-RUN tar -czf /build/libbpf.tar.gz /usr/lib/libbpf.a /usr/lib/pkgconfig/libbpf.pc /usr/include/bpf
+RUN mkdir /build && \
+    git clone --branch v1.0.1 --depth 1 https://github.com/libbpf/libbpf.git /build/libbpf && \
+    make -j $(nproc) -C /build/libbpf/src BUILD_STATIC_ONLY=y LIBSUBDIR=lib install && \
+    tar -czf /build/libbpf.tar.gz /usr/lib/libbpf.a /usr/lib/pkgconfig/libbpf.pc /usr/include/bpf
 
 COPY ./ /build/ebpf_exporter
 
 RUN cd /build/ebpf_exporter && \
-    PATH="/usr/lib/go-1.19/bin:$PATH" make build && \
+    make build && \
     /build/ebpf_exporter/ebpf_exporter --version
 
-FROM scratch as ebpf_exporter
+FROM gcr.io/distroless/static-debian11 as ebpf_exporter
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,27 +1,20 @@
-FROM ubuntu:22.04 as builder
+FROM golang:1.19-bullseye as builder
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends curl git ca-certificates build-essential libelf-dev gnupg2
+    apt-get install -y --no-install-recommends build-essential libelf-dev gnupg2
```
bobrik, line 4: With `golang:1.19-bullseye` you don't need:

* `build-essential` as we have all we need in the base image
* `gnupg2` as we don't add any new repos with their own keys

It works with just `libelf-dev`, so let's only keep that.

Dockerfile
```diff
@@ -1,27 +1,20 @@
...
 
-RUN echo 'deb https://ppa.launchpadcontent.net/longsleep/golang-backports/ubuntu jammy main' > /etc/apt/sources.list.d/golang-backports.list && \
-    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 52b59b1571a79dbc054901c0f6bc817356a3d45e && \
-    apt-get update && \
-    apt-get install -y --no-install-recommends golang-1.19-go
-
-RUN mkdir /build
```
bobrik, line 11: The number of layers doesn't really matter here as we only use the first image for building. None of the layers make it into the final image.

That said, the build steps are pretty fast and your grouping seems logical.

Dockerfile
```diff
@@ -1,27 +1,20 @@
...
-
-RUN git clone --branch v1.0.1 --depth 1 https://github.com/libbpf/libbpf.git /build/libbpf && \
-    make -C /build/libbpf/src BUILD_STATIC_ONLY=y LIBSUBDIR=lib install
-
-RUN tar -czf /build/libbpf.tar.gz /usr/lib/libbpf.a /usr/lib/pkgconfig/libbpf.pc /usr/include/bpf
+RUN mkdir /build && \
+    git clone --branch v1.0.1 --depth 1 https://github.com/libbpf/libbpf.git /build/libbpf && \
+    make -C /build/libbpf/src BUILD_STATIC_ONLY=y LIBSUBDIR=lib install && \
```
bobrik, line 8: Let's add `-j $(nproc)` before `-C` to make it build in parallel.

Dockerfile
```diff
@@ -1,27 +1,20 @@
...
 
 COPY ./ /build/ebpf_exporter
 
 RUN cd /build/ebpf_exporter && \
-    PATH="/usr/lib/go-1.19/bin:$PATH" make build && \
+    make build && \
     /build/ebpf_exporter/ebpf_exporter --version
 
-FROM scratch as ebpf_exporter
+FROM gcr.io/distroless/static-debian11 as ebpf_exporter
```
bobrik, line 17: What does this give us that `scratch` doesn't?

SuperQ: Mostly the CA certs, which can be used to do cert verification. I was thinking about adding the Prometheus exporter-toolkit which supports TLS and basic auth.

Dockerfile
SuperQ: Done.

Dockerfile
SuperQ: Done.

Dockerfile
bobrik: I'm not sure if anyone using TLS auth with a client cert signed by a public CA. My understanding is that people generally have a private CA and you need to supply the CA cert in that case, so having a public bundle available doesn't do you any good.

I searched a bit and it looks like you can use Let's Encrypt certs:.

* https://community.letsencrypt.org/t/user-client-certificates/100257/3

It doesn't look like you can request a specific subject field or anything like that with the config:

* https://github.com/prometheus/exporter-toolkit/blob/v0.1.0/https/README.md
* https://pkg.go.dev/crypto/tls#ClientAuthType

Which doesn't make it very secure with a public CA, as anyone with any cert from Let's Encrypt would be able to get access.

Given that, I think it might be good to in fact not have any trusted bundle in the docker image.

Let me know if I'm missing something here.

SuperQ: SAN verification for mTLS is on our TODO list. For now you have to use TLS and basic auth. Adding CA certs is a good step in the right direction.

bobrik: Do you expect people to use client certs signed by public CA? That's the only thing I can think of that requires public CA certs to be included.

SuperQ: I expect people to use a variety of certs. Publicly signed or not. This is an additive improvement, and it's useful to have certs here.

PR #159, opened 2022-10-23
Introduce post_attach_mark to allow initialization of metrics

bobrik:


SuperQ (approved):
Seems legit, init metrics is a good idea.

PR #160, opened 2022-10-23
Log startup time

bobrik:


cmd/ebpf_exporter/main.go
```diff
@@ -34,13 +37,14 @@ func main() {
...
 	}
+
 	configPath := filepath.Dir((*configFile).Name())
 	err = e.Attach(configPath)
 	if err != nil {
 		log.Fatalf("Error attaching exporter: %s", err)
 	}
 
-	log.Printf("Starting with %d programs found in the config", len(config.Programs))
+	log.Printf("Started with %d programs found in the config in %dms", len(config.Programs), time.Since(started).Milliseconds())
```
SuperQ, line 47: Do you know what steps are slow? Maybe each of the expensive startup steps should log their individual timings.

bobrik: Nothing is really slow today (especially compared to BCC when we needed to compile stuff), but I think it's good to have an indication of how fast things are in the logs.

With `timers`:

```
2022/10/23 19:08:26 Started with 1 programs found in the config in 6ms
```

With `bpf-jit`:

```
2022/10/23 19:08:18 Started with 1 programs found in the config in 58ms
```

SuperQ: Those timings are only once at startup? That doesn't seem like a highly useful thing to measure.

bobrik: Yes, only at startup. I find it useful.

Issue #143, opened 2022-10-19
Support fentry/fexit to replace kprobe/kretprobe


wenlxie:
We have a requirement that need to use fentry/fexit as much as possible to replace kprobe, 
But some of our nodes' kernel are still in old kernel and not support fentry. 
That means we need to solve following issues:
1. In node with old kernel, we need to use kprobe
2. In node with new kernel, we can use fentry to replace kprobe
3. Pods runs on these nodes with the same pod spec. 

One of the solution I am thinking is:
1. Add a kernel version limitation to the programs, if current program not met the requirements, then it will not be loaded
2. We will provide two programs for the same requirements but with different implementation, one is by fentry and another is kprobe. Only one will be loaded because of kernel version limitation configured 

@bobrik  Any suggestion for this? 
Thanks for your time.

bobrik:
It looks like `fentry` is also not yet available on arm64 even in 6.1-rc1. This patch is needed:

* https://patchwork.kernel.org/project/netdevbpf/cover/20220913162732.163631-1-xukuohai@huaweicloud.com/

I added `fentry` benchmarks in #144 and it looks much better than `kprobe` (as expected).

@zuzzas has #136 with both graceful handling for programs that cannot be loaded and kernel constraints.

I'll try to find some time to think how to make this work with the least amount of friction.

wenlxie:
The performance of `fentry` looks really amazing
@bobrik Many thanks for the infos. 
Kernel constraints  is a common requirements in Cloud.  Patiently wait  for the solutions.  Thanks

bobrik:
Is having the same config across the whole fleet the right idea to begin with?

What if somebody wants to load probes based on hardware, like NICs? What if there's something CPU model specific? We can't keep adding checks to ebpf_exporter to accommodate every situation.

I think there's a natural fit for external configuration management here. For physical hardware it might be whatever you use to configure the rest of it (we use Salt). For k8s you can write some sort of script that would do the equivalent work of checking any conditions you want for every program you have.

It can be as simple and as customizable as a python script:

```python
#!/usr/bin/env python3

import yaml
import platform
import packaging.version


def min_kernel_version(min_version):
    min_version = packaging.version.parse(min_version)
    running_version = packaging.version.parse(platform.release().split("-")[0])
    return running_version >= min_version


PROGRAMS = {
    "timers": [],
    "cgroup": [min_kernel_version("5.15")],
}

if __name__ == "__main__":
    enabled = []
[...]

PR #161, opened 2022-10-23
Enable loose attachment of programs

bobrik:
Now that we're using libbpf to drive attachment configuration from bpf code itself, it's no longer an option to only attach some of the programs.

In cases like cachestat where some functions might not be available for tracing, this is a bit annoying to work around.

Previously if you had a missing attach target, we would've just failed to start:

```
2022/10/23 16:44:21 Error attaching exporter: failed to attach to program "cachestat": failed to attach program "folio_account_dirtied": failed to attach program: no such file or directory
```

With the new code you would get a log line at startup:

```
2022/10/23 16:42:37 Failed to attach program "folio_account_dirtied" for "cachestat": failed to attach program: no such file or directory
```

And there would be an indication in metrics, allowing you to alert:

```
ebpf_exporter_ebpf_programs{function="add_to_page_cache_lru",program="cachestat",tag="6c007da3187b5b32"} 1
ebpf_exporter_ebpf_programs{function="folio_account_dirtied",program="cachestat",tag="6c007da3187b5b32"} 0
ebpf_exporter_ebpf_programs{function="mark_buffer_dirty",program="cachestat",tag="6c007da3187b5b32"} 1
ebpf_exporter_ebpf_programs{function="mark_page_accessed",program="cachestat",tag="6c007da3187b5b32"} 1
```

This might work fine for some programs, but not so well for others. Having an ability to alert makes it universally usable.

There's an option to add a flag to control this, we might do it later if there's any real world need for that.

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -29,7 +29,10 @@ type Exporter struct {
 	kaddrs                   map[string]uint64
 	enabledProgramsDesc      *prometheus.Desc
 	programInfoDesc          *prometheus.Desc
-	attachments              map[string][]progAttachment
+	programAttachedDesc      *prometheus.Desc
+	programRunTimeDesc       *prometheus.Desc
+	programRunCountDesc      *prometheus.Desc
+	attachedProgs            map[string]map[*libbpfgo.BPFProg]bool
 	descs                    map[string]map[string]*prometheus.Desc
 	decoders                 *decoder.Set
 }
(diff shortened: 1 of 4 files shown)


bobrik:
Same as before, we still recommend relying on external config generation with different programs if you want full control:

* https://github.com/cloudflare/ebpf_exporter/issues/143#issuecomment-1284832706

SuperQ:
Hmm, usually I would recommend having the attachment and info metrics be separated.

```
ebpf_exporter_ebpf_program_info
ebpf_exporter_ebpf_program_attached
```

When using info metrics, it's typical to do something like this:

```
some_metric * on (...) group_left (...) some_metric_info
```

This is why "info" metrics always have a value of 1. So that the multiply doesn't change the value. If it can vary, users would get unexpected results.

bobrik:
That's a good point, I'll update the PR.

bobrik:
Updated with separated metrics and additional run time metrics (if `kernel.bpf_stats_enabled` sysctl permits).

exporter/exporter.go
```diff
@@ -49,9 +52,30 @@ func New(cfg config.Config) (*Exporter, error) {
...
-		prometheus.BuildFQName(prometheusNamespace, "", "ebpf_programs"),
+		prometheus.BuildFQName(prometheusNamespace, "", "ebpf_program_info"),
 		"Info about ebpf programs",
-		[]string{"program", "function", "tag"},
+		[]string{"program", "function", "tag", "id"},
+		nil,
+	)
+
+	programAttachedDesc := prometheus.NewDesc(
+		prometheus.BuildFQName(prometheusNamespace, "", "ebpf_program_attached"),
```
SuperQ, line 62: Minor nit, I wonder if `ebpf_program` is redundant. Could simplify to just `program`.

SuperQ (approved):
LGTM

exporter/exporter.go
bobrik: There's a wrinkle to this: `program` can mean both "ebpf function" (`struct bpf_program` and whatever comes from it) and "a collection of ebpf functions and metrics" (what we call "program" in config).

Internally it's a bit of a mess, but in metrics `program` and `ebpf_program` are different things.

Maybe this is a good time to rename the larger thing to something like "measurement" or "config".

SuperQ: Makes sense!

PR #170, opened 2022-11-04
Support output with ring buffer

yunwei37:
Hi!

These patches add BPF ring buffer support for ebpf exporter. See https://www.kernel.org/doc/html/latest/bpf/ringbuf.html for the reason to use ring buf instead of perf event array in new kernels:

>There are two distinctive motivators for this work, which are not satisfied by existing perf buffer, which prompted creation of a new ring buffer implementation.
>- more efficient memory utilization by sharing ring buffer across CPUs;
> - preserving ordering of events that happen sequentially in time, even across multiple CPUs (e.g., fork/exec/exit events for a task).
>
> These two problems are independent, but perf buffer fails to satisfy both. Both are a result of a choice to have per-CPU perf ring buffer. Both can be also solved by having an MPSC implementation of ring buffer. The ordering problem could technically be solved for perf buffer with some in-kernel counting, but given the first one requires an MPSC buffer, the same solution would solve the second problem automatically.

I also add an example using the ring buf output, modified from https://github.com/libbpf/libbpf-bootstrap/blob/master/examples/c/bootstrap.c and and https://github.com/iovisor/bcc/blob/master/libbpf-tools/execsnoop.c

The output would be like:
```console
$ curl localhost:9435/metrics
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102180",ppid="102177"} 1
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102181",ppid="102177"} 1
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102182",ppid="102177"} 1
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102183",ppid="102177"} 1
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102186",ppid="102177"} 1
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102188",ppid="102177"} 1
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102190",ppid="102177"} 1
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102192",ppid="102177"} 1
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102201",ppid="102199"} 1
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat",pid="102202",ppid="102199"} 1
```


config/config.go
```diff
@@ -38,6 +38,7 @@ type Counter struct {
 	Name           string        `yaml:"name"`
 	Help           string        `yaml:"help"`
 	PerfEventArray bool          `yaml:"perf_event_array"`
+	RingBuf        bool          `yaml:"ring_buf"`
 	FlushInterval  time.Duration `yaml:"flush_interval"`
```
yunwei37, line 42: In fact, I think this may not be a good idea to create another option alone with  PerfEventArray, as more map types may be introduced in the future.

I think there might be two ways to handle this issue better:

1. we can infer the map type directly from the bpf object, with the given map name. So we may remove this field in the config struct.
2. Add an enum type to represent map type.

But they may bring some API break changes, so I left it here...Which do you like?

bobrik: I think we should infer the type, but not from the name. Every map has a type:

* https://pkg.go.dev/github.com/aquasecurity/libbpfgo@v0.4.2-libbpf-1.0.1#BPFMap.Type

And we can check `MapType` for what we want:

* https://pkg.go.dev/github.com/aquasecurity/libbpfgo@v0.4.2-libbpf-1.0.1#MapType

There's `MapTypePerfEventArray ` and `MapTypeRingbuf` there already.

examples/execsnoop.bpf.c
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-3-Clause
+/* Copyright (c) 2020 Facebook */
```
bobrik, line 2: Where exactly is this from?

examples/execsnoop.yaml
```diff
@@ -0,0 +1,22 @@
+metrics:
+  counters:
+    - name: process_exec_total
+      help: Count process starts
+      ring_buf: true
+      labels:
+        - name: pid
```
bobrik, line 7: I'm not sure this is a good example to have. Is there a more real world example?

bobrik (changes requested):

examples/execsnoop.bpf.c
yunwei37: From [bootstrap](https://github.com/libbpf/libbpf-bootstrap#bootstrap) in libbpf-bootstrap, and [execsnoop](https://github.com/iovisor/bcc/blob/master/libbpf-tools/execsnoop.bpf.c) from bcc/libbpf-tools.

This header declare is from https://github.com/libbpf/libbpf-bootstrap/blob/master/examples/c/bootstrap.bpf.c.

examples/execsnoop.yaml
yunwei37: Well, I'm not sure which example should I chose as well...

In fact all libbpf-tools with perf event array output can be converted in to ring buffer, they are already doing this: https://github.com/iovisor/bcc/pull/4262

Maybe I can try with other tools in [bcc/libbpf-tools](https://github.com/iovisor/bcc/blob/master/libbpf-tools)? For example:

- [sigsnoop](https://github.com/iovisor/bcc/blob/master/libbpf-tools/sigsnoop.bpf.c)
- [opensnoop](https://github.com/iovisor/bcc/blob/master/libbpf-tools/opensnoop.bpf.c)
- [mountsnoop](https://github.com/iovisor/bcc/blob/master/libbpf-tools/mountsnoop.bpf.c)
- [statsnoop](https://github.com/iovisor/bcc/blob/master/libbpf-tools/statsnoop.bpf.c)

Which one do you think is better?

yunwei37: Or maybe remove the `pid` and `ppid` will make the result from execsnoop more meaningful?

```
# TYPE ebpf_exporter_process_exec_total counter
ebpf_exporter_process_exec_total{comm="askpass.sh",filename="/home/yunwei/.vscode-server/bin/8fa188b2b301d36553cbc9ce1b0a146ccb93351f/extensions/git/dist/askpass.sh"} 2
ebpf_exporter_process_exec_total{comm="cat",filename="/usr/bin/cat"} 58
ebpf_exporter_process_exec_total{comm="cpuUsage.sh",filename="/home/yunwei/.vscode-server/bin/8fa188b2b301d36553cbc9ce1b0a146ccb93351f/out/vs/base/node/cpuUsage.sh"} 12
ebpf_exporter_process_exec_total{comm="curl",filename="/usr/bin/curl"} 3
ebpf_exporter_process_exec_total{comm="git",filename="/usr/bin/git"} 22
ebpf_exporter_process_exec_total{comm="git",filename="/usr/lib/git-core/git"} 5
ebpf_exporter_process_exec_total{comm="git-remote-http",filename="/usr/lib/git-core/git-remote-https"} 2
ebpf_exporter_process_exec_total{comm="mktemp",filename="/usr/bin/mktemp"} 2
ebpf_exporter_process_exec_total{comm="node",filename="/home/yunwei/.vscode-server/bin/8fa188b2b301d36553cbc9ce1b0a146ccb93351f/node"} 3
ebpf_exporter_process_exec_total{comm="ps",filename="/usr/bin/ps"} 12
ebpf_exporter_process_exec_total{comm="rm",filename="/usr/bin/rm"} 2
ebpf_exporter_process_exec_total{comm="sed",filename="/usr/bin/sed"} 24
```

It will shows the most frequently used commands or processed in the system

bobrik:
I reorganized output sink code a bit in a commit in my branch:

* https://github.com/bobrik/ebpf_exporter/commit/321f6d26a3f95d2ba155435c2fd263a858f0537a

I can't push it into this PR for some reason, you probably disabled that from your end.

That said, I'm no longer sure that we have a use case for either `perf_buf` or `ring_buf`.

One can accomplish limited cardinality with `BPF_MAP_TYPE_LRU_HASH`:

```c
struct {
    __uint(type, BPF_MAP_TYPE_LRU_HASH);
    __uint(max_entries, 10);
    __type(key, u64);
    __type(value, u64);
} oom_kills_total SEC(".maps");
```

We can have periodic metric flush generalized too. I am not yet sure if that is necessary.

@zuzzas, @yunwei37, can you think of an example where that is no sufficient and we need to pass individual events?

yunwei37:
> I can't push it into this PR for some reason, you probably disabled that from your end.

I have enabled `Allow edits by maintainers`, maybe I need to do something else?

> can you think of an example where that is no sufficient and we need to pass individual events?

I can not find an example that it's necessary for us to pass individual events in the exporter now, maybe someone can think it later?

However, for keep consist with https://github.com/iovisor/bcc/tree/master/libbpf-tools, I suggest keep `perf_buf` or `ring_buf` both as an option? A number of tools in libbpf-tools use `perf_buf` or `ring_buf` to pass individual events to userspace, so keep these map types will not require users to rewrite the output of these tools with `BPF_MAP_TYPE_LRU_HASH`.

bobrik:
`ebpf_exporter` is not intended for "snoop" type tools, it's for metrics. If you want to pull individual events from the kernel, you probably want to push it into some sort of logging system, not into metrics that are scraped at regular intervals.

yunwei37:
I see. Thank you very much!

bobrik:
I tried converting oomkill example to `BPF_MAP_TYPE_LRU_HASH` and it didn't work well on the decoding side. The reason is that we decode at scrape and by that time the cgroup is gone, so we end up with `unknown_cgroup_id:XYZ`, which is not great. With the existing perf event array setup the event is read and decoded while cgroup is still present, so it can be decoded properly. It works mostly because of a race condition (systemd can in theory remove cgroups faster that we decode the event), but it does work.

This feels kind of cgroup specific, so maybe we should just use inotify to keep the cgroup mapping up to date.

bobrik:
I submitted `fanotify` support for cgroup fs upstream and opened a PR to make use of it: #244.

PR #177, opened 2023-01-09
Update module path to adhere with Go versioning guidelines

epk:
Signed-off-by: Aditya Sharma <aditya.sharma@shopify.com>

## Context

When trying to install/import `ebpf_exporter@v2.0.0` as a Go dependency, the toolchain throws the following error.

```
go install github.com/cloudflare/ebpf_exporter/cmd/ebpf_exporter@v2.0.0
go: github.com/cloudflare/ebpf_exporter/cmd/ebpf_exporter@v2.0.0: github.com/cloudflare/ebpf_exporter@v2.0.0: invalid version: module contains a go.mod file, so module path must match major version ("github.com/cloudflare/ebpf_exporter/v2")
```
OR
```
go get github.com/cloudflare/ebpf_exporter@v2.0.0
go: github.com/cloudflare/ebpf_exporter@v2.0.0: invalid version: module contains a go.mod file, so module path must match major version ("github.com/cloudflare/ebpf_exporter/v2")
```

## Fix

Updated the module paths to include `/v2` by using https://github.com/icholy/gomajor:
```bash
gomajor path -next
```

cc @bobrik

### DIFF
--- a/benchmark/getpid_test.go
+++ b/benchmark/getpid_test.go
@@ -7,7 +7,7 @@ import (
 	"unsafe"
 
 	"github.com/aquasecurity/libbpfgo"
-	"github.com/cloudflare/ebpf_exporter/util"
+	"github.com/cloudflare/ebpf_exporter/v2/util"
 )
 
 func BenchmarkGetpidWithoutAnyProbes(b *testing.B) {
--- a/cmd/ebpf_exporter/main.go
+++ b/cmd/ebpf_exporter/main.go
@@ -6,8 +6,8 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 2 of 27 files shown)


bobrik:
Makes sense, thanks! What are you embedding `ebpf_exporter` into?

epk:
> Makes sense, thanks! What are you embedding `ebpf_exporter` into?

Instead of having a fork just for building a docker image, we declare `github.com/cloudflare/ebpf_exporter/v2/cmd/ebpf_exporter` as a Go dependency[^1] and build it alongside our eBPF probes.



[^1]: https://github.com/go-modules-by-example/index/blob/ac9bf72/010_tools/README.md

PR #179, opened 2023-01-18
fix defer in a for loop function.

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

Do not use defer in a loop unless you’re sure about what you’re doing. It may not work as expected.

[#2 — Defer inside a loop ](https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01)


bobrik:
Unless you have some specific issue, it does work as expected here.

yanggangtony:
> Unless you have some specific issue, it does work as expected here.

Nop..
I just saw it on golang-practise ..
Please feel good to close it..

Issue #208, opened 2023-05-26
XDP Support


huseyinsaatci:
I was using ebpf_exporter in my project and realized that there is no XDP support in the project. I forked the project and edited it to support XDP programs. I used libbpfgo's _AttachXDP_ function to attach the XDP program to the required interface.

My edit contains two new fields in YAML files: **is_xdp**  to indicate if the metric has an XDP program, and **interfaces** is an array with interfaces to attach the XDP program. And usage is like:
```
is_xdp: true
interfaces: ["enp8s0", "docker0"]
```
If you consider it worth adding support for XDP programs I can open a PR.

bobrik:
What do you think about using `SEC("xdp/enp8s0")` instead (see: https://github.com/cloudflare/ebpf_exporter/pull/154)? The downside is that the interface is hardcoded in a program, but the upside is that it's consistent with other `SEC` setups.

In your setup I don't think there's a need for `is_xdp`. You can detect it from `SEC("xdp")`. Having `interfaces` set in the config means that you attach all programs to all interfaces. I can imagine cases where you'd want to have different programs on different interfaces as a part of the same config.

huseyinsaatci:
Makes sense but when I try to use SEC("xdp/enp8s0") in my program I get this error: 
```
missing BPF prog type, check ELF section name 'xdp/enp8s0
failed to load: -22
Error attaching exporter: error loading bpf object from "src/xdp.bpf.o" for config "xdp": failed to load BPF object: invalid argument'
```
I don't think it recognizes `'xdp/enp8s0'` as an XDP program.
It is possible to get the interface from the SEC part but only if this problem can be solved.

bobrik:
I linked https://github.com/cloudflare/ebpf_exporter/pull/154 because it implemented `SEC("perf_event/type=0,config=3,frequency=99")`.

You can do a similar thing here to implement `SEC("xdp/enp8s0")`.

huseyinsaatci:
Okay, I will examine the codes and try to implement a similar one. I hope it's not a problem if I ask questions about the codes if I have problems.

bobrik:
I merged #215, thanks!

PR #215, opened 2023-06-26
Implementation of XDP support

huseyinsaatci:
Implementation of XDP support for issue #208.

### DIFF
--- a/examples/xdp.bpf.c
+++ b/examples/xdp.bpf.c
@@ -45,8 +45,7 @@ parse_header(ipv6hdr)
 parse_header(tcphdr)
 parse_header(udphdr)
 
-SEC("xdp/lo")
-int xdp_trace(struct xdp_md *ctx) {
+static int xdp_trace(struct xdp_md *ctx) {
     void *data_end = (void *) (long) ctx->data_end;
     void *data = (void *) (long) ctx->data;
     struct packet_key_t key = {};
@@ -105,6 +104,11 @@ int xdp_trace(struct xdp_md *ctx) {
     increment_map(&xdp_incoming_packets_total, &key, 1);
 
     return XDP_PASS;
+}
+
+SEC("xdp/lo")
+int trace_lo(struct xdp_md *ctx) {
+    return xdp_trace(ctx);
 }
 
 char LICENSE[] SEC("license") = "GPL";
--- a/examples/xdp.bpf.c
+++ b/examples/xdp.bpf.c
@@ -3,119 +3,108 @@
 #include <bpf/bpf_endian.h>
 #include "maps.bpf.h"
 
-char LICENSE[] SEC("license") = "Dual BSD/GPL";
-
 #define ETH_P_IPV6 0x86DD
 #define ETH_P_IP 0x0800
-#define ETH_P_ARP 0x0806
 
-struct tcp_udp_out
-{
-  u32 src_ip;
-  u32 dst_ip;
-  u32 src_port;
-  u32 dst_port;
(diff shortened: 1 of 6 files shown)


bobrik:
I updated the PR to make it match the rest of the repo a bit better.

The biggest change is the example, where consolidated multiple metrics into one, added IPv6 support and made it's cardinality lower by not including actual addresses.

Here's how the metric loos like now:

```
# HELP ebpf_exporter_xdp_incoming_packets_total Incoming packets going through xdp by protocol and port
# TYPE ebpf_exporter_xdp_incoming_packets_total counter
ebpf_exporter_xdp_incoming_packets_total{dst_port="0",ip="ipv4",protocol="icmp"} 5
ebpf_exporter_xdp_incoming_packets_total{dst_port="53",ip="ipv4",protocol="udp"} 3
ebpf_exporter_xdp_incoming_packets_total{dst_port="9435",ip="ipv4",protocol="tcp"} 3
ebpf_exporter_xdp_incoming_packets_total{dst_port="9435",ip="ipv6",protocol="tcp"} 7
```

Let me know if it looks good to you.

huseyinsaatci:
It seems all great but I just have one question.
If I define a SEC part that has one invalid interface before the others like `SEC("xdp/enp8s0,asdsd,lo")`, the others won't be attached because of the early error. In this scenario, `enp8s0` and `lo` are valid, `enp8s0` will be attached but `lo` won't be because of the invalid one `asdsd`. Would not it be better if we solved this problem and attach all valid interfaces?

bobrik:
That's a good concern.

I think we shouldn't allow attaching to multiple interfaces in one `SEC`, because it makes `xdp` special.

For any other program we would export a metric like this when it fails to attach:

```
# HELP ebpf_exporter_ebpf_program_attached Whether a program is attached
# TYPE ebpf_exporter_ebpf_program_attached gauge
ebpf_exporter_ebpf_program_attached{id="329"} 0

# HELP ebpf_exporter_ebpf_program_info Info about ebpf programs
# TYPE ebpf_exporter_ebpf_program_info gauge
ebpf_exporter_ebpf_program_info{config="something",id="329",program="trace_something",tag="bd45628e0edfd896"} 1
```

For `xdp` we can have partial failures and that's not great.

I added a commit to only allow a single interface in `SEC("xdp")`. This way a user can have multiple `SEC`s:

```c
SEC("xdp/lo")
int trace_lo(struct xdp_md *ctx) {
    return xdp_trace(ctx);
}

SEC("xdp/lol")
int trace_lol(struct xdp_md *ctx) {
    return xdp_trace(ctx);
}
```

And they will know which one of them fails to attach, if any:

```
# HELP ebpf_exporter_ebpf_program_attached Whether a program is attached
# TYPE ebpf_exporter_ebpf_program_attached gauge
ebpf_exporter_ebpf_program_attached{id="329"} 1
ebpf_exporter_ebpf_program_attached{id="330"} 0

# HELP ebpf_exporter_ebpf_program_info Info about ebpf programs
# TYPE ebpf_exporter_ebpf_program_info gauge
ebpf_exporter_ebpf_program_info{config="xdp",id="329",program="trace_lo",tag="bd45628e0edfd896"} 1
ebpf_exporter_ebpf_program_info{config="xdp",id="330",program="trace_eth0",tag="bd45628e0edfd896"} 0
```

It's less concise, but it's consistent with how other things work.

Does this make sense?

huseyinsaatci:
It's clearer to separate interfaces when I think it this way. It seems all good, we can roll it out.

PR #235, opened 2023-07-18
Add proto and port breakdown to kfree_skb

bobrik:
Example metrics:

    # HELP ebpf_exporter_kfree_skb_total Timers number of calls to kfree_skb
    # TYPE ebpf_exporter_kfree_skb_total counter
    ebpf_exporter_kfree_skb_total{eth_proto="ipv4",ip_proto="tcp",port="38188",reason="SKB_DROP_REASON_NOT_SPECIFIED"} 3
    ebpf_exporter_kfree_skb_total{eth_proto="ipv4",ip_proto="tcp",port="38188",reason="SKB_DROP_REASON_NO_SOCKET"} 2
    ebpf_exporter_kfree_skb_total{eth_proto="ipv4",ip_proto="tcp",port="43536",reason="SKB_DROP_REASON_NOT_SPECIFIED"} 1
    ebpf_exporter_kfree_skb_total{eth_proto="ipv4",ip_proto="tcp",port="43548",reason="SKB_DROP_REASON_NOT_SPECIFIED"} 1
    ebpf_exporter_kfree_skb_total{eth_proto="ipv4",ip_proto="tcp",port="43560",reason="SKB_DROP_REASON_NOT_SPECIFIED"} 1
    ebpf_exporter_kfree_skb_total{eth_proto="ipv6",ip_proto="udp",port="12345",reason="SKB_DROP_REASON_SOCKET_RCVBUFF"} 448205

### DIFF
--- a/examples/kfree_skb.bpf.c
+++ b/examples/kfree_skb.bpf.c
@@ -1,19 +1,98 @@
 #include <vmlinux.h>
 #include <bpf/bpf_tracing.h>
+#include <bpf/bpf_endian.h>
 #include "maps.bpf.h"
+
+#define ETH_P_IPV6 0x86DD
+#define ETH_P_IP 0x0800
+
+struct kfree_skb_key_t {
+    u16 eth_proto;
+    u16 ip_proto;
+    u16 port;
+    u16 reason;
+};
 
 struct {
     __uint(type, BPF_MAP_TYPE_HASH);
-    __uint(max_entries, 1024);
-    __type(key, u16);
+    __uint(max_entries, 10240);
+    __type(key, struct kfree_skb_key_t);
     __type(value, u64);
 } kfree_skb_total SEC(".maps");
 
-SEC("tp_btf/kfree_skb")
-int BPF_PROG(kfree_skb, struct sk_buff *skb, void *location, enum skb_drop_reason reason)
+SEC("tp_btf/kfree_skb") int BPF_PROG(kfree_skb, struct sk_buff *skb, void *location, enum skb_drop_reason reason)
 {
-    u16 key = reason;
+    struct kfree_skb_key_t key;
+    struct ethhdr eth_hdr;
+    struct iphdr ip_hdr;
+    struct ipv6hdr ipv6_hdr;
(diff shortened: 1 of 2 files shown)


examples/kfree_skb.bpf.c
```diff
@@ -1,19 +1,98 @@
...
+        if (bpf_probe_read_kernel(&tcp_hdr, sizeof(tcp_hdr), skb->head + skb->transport_header) < 0) {
+            return 0;
+        }
+        key.port = bpf_ntohs(tcp_hdr.dest);
+        break;
+    case IPPROTO_UDP:
+        if (bpf_probe_read_kernel(&udp_hdr, sizeof(udp_hdr), skb->head + skb->transport_header) < 0) {
+            return 0;
+        }
+        key.port = bpf_ntohs(udp_hdr.dest);
```
bobrik, line 88: It is unclear how to connect this to a socket that is bound locally. It might be easier if filtered by `reason`.

In #221 it's trivial because we have a `struct sock` that relates to a local listening socket rather than a more abstract `struct skb` that can be for any network packet.

bobrik:
This setup is a lot more complex to accomplish what #221 does easily. It also requires the right `reason` to be set when listen drops happen, which is not the case in v6.5 at the time of writing.

The tracepoint would fire a lot more often too, resulting in more overhead.

examples/kfree_skb.bpf.c
```diff
@@ -1,19 +1,98 @@
...
+        return 0;
+    }
+
+    key.eth_proto = bpf_ntohs(eth_hdr.h_proto);
+
+    if (!key.eth_proto && !bpf_ntohs(skb->protocol)) {
+        return 0;
+    }
+
+    switch (key.eth_proto) {
```
bobrik, line 49: Do we have to be vlan aware here? What about other encapsulation?

See: https://github.com/dsahern/bpf-progs/blob/175e2dfd/src/parse_pkt.c#L167-L179

bobrik:
Overhead of calling into a tracepoint that does nothing seems to be ~15ns according to benchmarks from #236.

PR #251, opened 2023-08-20
Add icmp-ip example with inet_ip decoder

bobrik:
Example metrics from local pings:

    ivan@vm:~$ curl -s http://localhost:9435/metrics | fgrep ebpf_exporter_icmp
    # HELP ebpf_exporter_icmp4_received_packets_total Number of icmp packets received over IPv4
    # TYPE ebpf_exporter_icmp4_received_packets_total counter
    ebpf_exporter_icmp4_received_packets_total{source_addr="10.2.0.15"} 4
    ebpf_exporter_icmp4_received_packets_total{source_addr="127.0.0.1"} 6
    # HELP ebpf_exporter_icmp6_received_packets_total Number of icmp packets received over IPv6
    # TYPE ebpf_exporter_icmp6_received_packets_total counter
    ebpf_exporter_icmp6_received_packets_total{source_addr="::1"} 4
    ebpf_exporter_icmp6_received_packets_total{source_addr="fe80::5054:ff:fe12:3456"} 4
    ebpf_exporter_icmp6_received_packets_total{source_addr="fec0::5054:ff:fe12:3456"} 6

PR #254, opened 2023-08-28
Add an eBPF program to measure synchronous connect() calls latencies

alexclear:


examples/connect-latency.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+};
+
+struct connect_latency_key_t {
+    u32 d_ip;    // Destination IPv4 address
+    u16 d_port;  // Destination port number
+    u64 slot;
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
```
bobrik, line 28: `BPF_MAP_TYPE_LRU_HASH` is preferred for resiliency against map getting full.

examples/connect-latency.yaml
```diff
@@ -0,0 +1,21 @@
+metrics:
+  histograms:
+    - name: connect_latency_seconds
+      help: Latency histogram for TCP connect() syscall
```
bobrik, line 4: If it's TCP only, then let's call the metric `tcp_connect_latency_seconds`  and rename the file into `tcp-connect-latency.yaml`.

examples/connect-latency.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+} connect_start SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, MAX_LATENCY_SLOT + 2);
+    __type(key, struct connect_latency_key_t);
+    __type(value, u64);
+} connect_latency_seconds SEC(".maps");
+
+static inline __u16 ntohs(__u16 value) {
```
bobrik, line 41: Any reason not to use `bpf_ntohs`?

examples/connect-latency.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+
+SEC("kprobe/__sys_connect")
+int BPF_KPROBE(kprobe__sys_connect, int sockfd, const struct sockaddr *addr, int addrlen)
+{
+    struct sockaddr sa;
+    struct connect_start_val_t start_val = {};
+
+    bpf_probe_read(&sa, sizeof(sa), addr);
+
+    if (sa.sa_family == 1) {
```
bobrik, line 53: There are lots of families and you probably only care about `AF_INET`:

* https://elixir.bootlin.com/linux/v6.6-rc1/source/include/linux/socket.h#L192

I suggest you add `#define AF_INET 2` and use it here.

examples/connect-latency.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+    __uint(max_entries, MAX_LATENCY_SLOT + 2);
+    __type(key, struct connect_latency_key_t);
+    __type(value, u64);
+} connect_latency_seconds SEC(".maps");
+
+static inline __u16 ntohs(__u16 value) {
+    return ((value & 0x00FF) << 8) | ((value & 0xFF00) >> 8);
+}
+
+SEC("kprobe/__sys_connect")
```
bobrik, line 45: `fentry` is a lot faster:

* https://github.com/cloudflare/ebpf_exporter/tree/master/benchmark

You also get:

* Function arguments in `fexit`, allowing you to key on `addr` rather than pid
* No need for `BPF_CORE_READ` as BTF allows direct reads

examples/connect-latency.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+
+            start_val.d_ip = BPF_CORE_READ(&v6.sin6_addr.in6_u, u6_addr32[3]);
+            start_val.d_port = v6.sin6_port;
+        } else {
+            const char debug_str[] = "This is native ipv6, I'm giving up!\n";
+            bpf_trace_printk(debug_str, sizeof(debug_str));
+            return 0;
+        }
+    } else {
+        const char debug_str[] = "Unexpected addrlen: %d, address family: %d\n";
```
bobrik, line 80: We generally don't leave debug statements around.

examples/connect-latency.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+        bpf_probe_read(&v6, sizeof(v6), addr);
+
+        if (BPF_CORE_READ(&v6.sin6_addr.in6_u, u6_addr32[0]) == 0x00000000 &&
+            BPF_CORE_READ(&v6.sin6_addr.in6_u, u6_addr32[1]) == 0x00000000 &&
+            BPF_CORE_READ(&v6.sin6_addr.in6_u, u6_addr32[2]) == 0x0000FFFF) {
+
+            start_val.d_ip = BPF_CORE_READ(&v6.sin6_addr.in6_u, u6_addr32[3]);
+            start_val.d_port = v6.sin6_port;
+        } else {
+            const char debug_str[] = "This is native ipv6, I'm giving up!\n";
```
bobrik, line 75: Please implement IPv6 as well. Normally it's in a separate map: #251.

examples/connect-latency.yaml
bobrik: Since we do this:

```
// Filter out non-blocking sockets and errors
```

It's probably a good idea to add `blocking` in the name as well.

examples/connect-latency.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+        const char debug_str[] = "Did not find anything in the map!\n";
+        bpf_trace_printk(debug_str, sizeof(debug_str));
+        return 0;
+    }
+    if (ret != 0) {
+        return 0; // Filter out non-blocking sockets and errors
+    }
+    const char debug_str[] = "Return code is: %d\n";
+    bpf_trace_printk(debug_str, sizeof(debug_str), ret);
+    struct connect_latency_key_t key = {};
```
bobrik, line 112: Please put all definitions at the top of the function.

examples/connect-latency.bpf.c
bobrik: I would recommend `bpf_printk()` when you debug locally:

* https://nakryiko.com/posts/bpf-tips-printk/

examples/connect-latency.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+#define MAX_LATENCY_SLOT 26
+
+struct connect_start_key_t {
+    u64 pid_tgid;
+};
+
+struct connect_start_val_t {
+    u64 ts;
+    int addrlen;
+    u32 d_ip;    // Destination IPv4 address
```
bobrik, line 17: `clang-format` is not happy here

examples/connect-latency.bpf.c
```diff
@@ -0,0 +1,133 @@
...
+
+#define MAX_LATENCY_SLOT 26
+
+struct connect_start_key_t {
+    u64 pid_tgid;
+};
+
+struct connect_start_val_t {
+    u64 ts;
+    int addrlen;
```
bobrik, line 16: You aren't really using `addrlen`.

bobrik (changes requested):

Issue #253, opened 2023-08-22
how to support struct type value


cdkunsong:
hi,
I also want to discuss a issue, whether it is possible to use struct type for value,
although I see in tcp-window-clamps.bpf.c

struct socktest {
     u32 value1;
     u32 value2;
     u32 value3;
}
struct {
    __uint(type, BPF_MAP_TYPE_LRU_HASH);
    __uint(max_entries, 1024);
    __type(key, u64);
    __type(value, struct socktest *);
} tcp_rmem_schedule_enters SEC(".maps");

but it don't use by metrics .
Currently, it seems that a struct type cannot be supported, and it can parse multiple member values.

Can you explain this problem, thanks!

bobrik:
Please use [markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code) when you quote code, it's hard to read code without any indentation.

Prometheus metrics have basic types: counter, gauge, histograms (which is just a bunch of counters). All of these have just float64 as values, as described in prometheus data model:

* https://prometheus.io/docs/concepts/data_model/

Given that, I'm not sure what you would do with a struct in a map value.

ntnx-aleksa:
Hello,

could you consider allowing struct values and similarly having labels and decoders for each struct member? I think this is very useful. I am counting a large amount of values in a struct because they all belong to the same key. For ebpf_exporter I would have to split this struct into a dozen maps with value u64. This will probably have a lot of lookup overhead. 
If there is a better alternative or I'm misunderstanding something, please let me know. 

Thank you.

bobrik:
What is your concrete use-case? What kind of metrics and labels do you want to produce?

ntnx-aleksa:
Let's say I'm collecting a dozen counters, that is u64 monotonically increasing variables, all related to the same block device - it makes perfect sense to group these together. I have a per-cpu hash map with a `u32` key, decoded as `majorminor` and a `struct disk_counters` value type.

For example:
```
struct disk_counters {
    unsigned long counter;
    unsigned long another_counter;
    unsigned long yet_another_counter;
    // ...
}
```

Currently, I would have to split this struct by creating a new map for each field. I tried just that and profiling the two BPF programs with bpftool shows degraded performance on the ebpf_exporter-compatible program. 
[...]

PR #257, opened 2023-08-31
Validate expected value size for maps

bobrik:
See: https://github.com/cloudflare/ebpf_exporter/issues/253#issuecomment-1694207100


bobrik:
With u32:

```
2023/08/31 05:59:32 Error attaching exporter: error validating maps for config "pci": value size for map "pci_user_read_config_ops_total" is not expected 8 bytes (u64), it is 4 bytes
```

PR #276, opened 2023-09-19
Minor adjustsments

netoptimizer:
Hi @bobrik 

I made these two small adjustments while trying out the benchmarks in ebpf_exporter.

### DIFF
--- a/benchmark/probes/benchmark.bpf.h
+++ b/benchmark/probes/benchmark.bpf.h
@@ -3,7 +3,7 @@
 #include <bpf/bpf_tracing.h>
 
 #if defined(__TARGET_ARCH_x86)
-#define FENTRY_SEC() SEC("fentry/__x64_sys_getpid")
+#define FENTRY_SEC() SEC("fentry/__do_sys_getpid")
 #elif defined(__TARGET_ARCH_arm64)
 #define FENTRY_SEC() SEC("fentry/__arm64_sys_getpid")
 #else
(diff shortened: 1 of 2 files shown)


benchmark/probes/benchmark.bpf.h
```diff
@@ -3,15 +3,15 @@
 #include <bpf/bpf_tracing.h>
 
 #if defined(__TARGET_ARCH_x86)
-#define FENTRY_SEC() SEC("fentry/__x64_sys_getpid")
+#define FENTRY_SEC() SEC("fentry/__do_sys_getpid")
```
bobrik, line 6: I can see both on our x86_64 running v6.1:

```
$ cat /proc/kallsyms | grep -E '( __x64_sys_getpid|__do_sys_getpid)$'
0000000000000000 t __do_sys_getpid
0000000000000000 T __x64_sys_getpid
```

On arm64 we only have one:

```
$ cat /proc/kallsyms | grep -E '( __arm64_sys_getpid|__do_sys_getpid)$'
0000000000000000 T __arm64_sys_getpid
```

Switching to `__do_sys_getpid` makes sense given what you say, even if it makes me sad for the lack of consistency.

cgroup/fanotify.go
```diff
@@ -188,7 +188,8 @@ func (m *fanotifyMonitor) Resolve(id int) string {
 }
 
 // The following kernel patch is required to take advantage of this (included in v6.6-rc1):
-// * https://github.com/torvalds/linux/commit/0ce7c12e88cf
+// * https://git.kernel.org/torvalds/c/0ce7c12e88cf ("kernfs: attach uuid for every kernfs and report it in fsid")
+//
```
bobrik, line 192: Drop this line and LGTM. I think it might make the linter unhappy.

PR #299, opened 2023-10-16
Add org.opencontainers.image labels to the docker image

bobrik:


bobrik:
Turns out these are populated automatically, but Github is not picking them up. I also fixed a stray space at the beginning of the description.

PR #309, opened 2023-10-19
biolatency: Fix kernel version check.

ntnx-aleksa:
The tracepoint definition change in v5.11-rc1 mentioned in the biolatency example program seems to have been [backported into v5.10.137](https://elixir.bootlin.com/linux/v5.10.137/source/include/trace/events/block.h#L206).
This small fix allows the program to load and work on kernels between v5.10.137 and v5.11. I deem this PR small enough not to have a related issue opened.


bobrik:
This stuff was mostly copied from bcc, so you should probably make a PR there as well:

* https://github.com/iovisor/bcc/blob/v0.28.0/libbpf-tools/biolatency.bpf.c#L79

bobrik (approved):

bobrik:
Thanks!

Issue #321, opened 2023-11-16
validateMaps fails for eBPF map without property 'value'


havefuncoding1:
Almost all yaml files under examples point to eBPF maps with 'value' property defined. Here is an example from accept-latency:
```
struct {
    __uint(type, BPF_MAP_TYPE_HASH);
    __uint(max_entries, (MAX_LATENCY_SLOT + 1) * MAX_PORTS);
    __type(key, struct socket_latency_key_t);
    __type(value, u64);
} accept_latency_seconds SEC(".maps");
```
One exception is that oomkill. Here is the map definition which has no 'value'. Instead, it is value_size.
```
struct {
    __uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
    __uint(key_size, sizeof(u32));
    __uint(value_size, sizeof(u32));
} oom_kills_total SEC(".maps");
```
Running this example results in runtime failure
```
Error attaching exporter: error validating maps for config "oomkill": value size for map "oom_kills_total" is not expected 8 bytes (u64), it is 4 bytes
```
And this is caused by the check in validateMap. What would be the solution here? Is replacing 'fmt.Errorf' with a warning message reasonable? We can add a 'value' to the map, but per my understanding there shouldn't be a 'value' property for BPF_MAP_TYPE_PERF_EVENT_ARRAY. [Here](BPF_MAP_TYPE_PERF_EVENT_ARRAY ) is reference from Nakryiko's blog. Thanks.
```
valueSize := m.ValueSize()
if valueSize != 8 {
	return fmt.Errorf("value size for map %q is not expected 8 bytes (u64), it is %d bytes", name, valueSize)
}
```

bobrik:
Thanks for the report. I opened #322 to both fix this and prevent it from breaking again.

PR #322, opened 2023-11-16
Run config check in CI

bobrik:


bobrik:
The second commit fixes #257, as reported in #321.

bobrik:
Hardening this in #326.

PR #320, opened 2023-11-14
Added support for external BTF

havefuncoding1:
1. Introduced new command line argument '--btfpath' to support external BTF so the tool can be used on system without vmlinx. External BTFs are available [here](https://github.com/aquasecurity/btfhub-archive).
2. Updated kernel version check for biolatency. Without it, in my environment running 5.10.178 kernel, the exporter runtime fails with 'Invalid argument'. @ntnx-aleksa just wanted to bring this to your awareness since you made initial update in commit [5511eae](5511eaeeea6745a2508e2786b586c865444a81f6). Hope changes here don't break anything.

PR #323, opened 2023-11-17
Enabled support for external BTF

bobrik:
This replaces #320.


bobrik:
Thanks @havefuncoding1!

PR #342, opened 2024-02-03
Add decoder cache

bobrik:
Two commits:

---

### Add decoder cache

When using percpu maps on large machines, one can find themself decoding the same set of inputs to the same set of values over and over again.

Given that the configuration is static and inputs generally do not disappear, it is a good idea to add a cache to speed up decoding.

Unfortunately, Go does not allow having `[]byte` as a map key, so we have to convert the cache key to a string, making it a little bit more expensive that it could've been otherwise.

Here's a benchmark:

```
goos: linux
goarch: arm64
pkg: github.com/cloudflare/ebpf_exporter/v2/decoder
BenchmarkCache/direct-10                 2328188               517.7 ns/op           264 B/op         17 allocs/op
BenchmarkCache/cached-10                23191784                48.9 ns/op            64 B/op          2 allocs/op
```

---

### Improve cache efficiency in decoder

Instead of converting `[]byte` to `string` to use it in a map key, we can hash `[]byte` to use as a key and make sure that there are no collisions as a result of this. This is quite a bit faster (before and after):

```
BenchmarkCache/cached-10             23694093            49.04 ns/op          64 B/op           2 allocs/op
BenchmarkCache/cached-10            132218238             9.29 ns/op           0 B/op           0 allocs/op
```

### DIFF
--- a/decoder/decoder.go
+++ b/decoder/decoder.go
@@ -1,10 +1,8 @@
 package decoder
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
-	"hash/maphash"
 	"sync"
 
 	"github.com/cloudflare/ebpf_exporter/v2/config"
@@ -13,11 +11,6 @@ import (
 
 // ErrSkipLabelSet instructs exporter to skip label set
 var ErrSkipLabelSet = errors.New("this label set should be skipped")
-
-type cacheValue struct {
-	in  []byte
-	out []string
-}
 
 // Decoder transforms byte field value into a byte value representing string
 // to either use as an input for another Decoder or to use as the final
@@ -30,8 +23,7 @@ type Decoder interface {
 type Set struct {
 	mu       sync.Mutex
 	decoders map[string]Decoder
-	cache    map[uint64]cacheValue
-	seed     maphash.Seed
+	cache    map[string][]string
 }
 
 // NewSet creates a Set with all known decoders
@@ -65,8 +57,7 @@ func NewSet() (*Set, error) {
 			"syscall":      &Syscall{},
 			"uint":         &UInt{},
 		},
-		cache: map[uint64]cacheValue{},
-		seed:  maphash.MakeSeed(),
+		cache: map[string][]string{},
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


decoder/decoder.go
```diff
@@ -87,6 +98,26 @@ func (s *Set) Decode(in []byte, label config.Label) ([]byte, error) {
 // DecodeLabels transforms eBPF map key bytes into a list of label values
 // according to configuration
 func (s *Set) DecodeLabels(in []byte, labels []config.Label) ([]string, error) {
+	cacheKey := maphash.Bytes(s.seed, in)
```
uhthomas, line 101: I mentioned this internally, but it would be simpler to use `map[string]cacheValue{}` and `s.cache[string(in)]`. The Go compiler has a specific optimisation for this case.

https://github.com/golang/go/issues/3512

uhthomas: current:

```sh
goos: linux
goarch: amd64
pkg: github.com/cloudflare/ebpf_exporter/v2/decoder
cpu: AMD Ryzen 9 3950X 16-Core Processor
BenchmarkCache/direct-32                 1097241              1155 ns/op
BenchmarkCache/cached-32                126773032                9.419 ns/op
```

with optimisation:

```sh
goos: linux
goarch: amd64
pkg: github.com/cloudflare/ebpf_exporter/v2/decoder
cpu: AMD Ryzen 9 3950X 16-Core Processor
BenchmarkCache/direct-32                 1102927              1032 ns/op
BenchmarkCache/cached-32                134451720                9.002 ns/op
```

```diff
diff --git a/decoder/decoder.go b/decoder/decoder.go
--- a/decoder/decoder.go
+++ b/decoder/decoder.go
@@ -1,7 +1,6 @@
 package decoder

 import (
-       "bytes"
        "errors"
        "fmt"
        "hash/maphash"
@@ -30,7 +29,7 @@
 type Set struct {
        mu       sync.Mutex
        decoders map[string]Decoder
-       cache    map[uint64]cacheValue
+       cache    map[string]cacheValue
        seed     maphash.Seed
 }

@@ -65,7 +64,7 @@
                        "syscall":      &Syscall{},
                        "uint":         &UInt{},
                },
-               cache: map[uint64]cacheValue{},
+               cache: map[string]cacheValue{},
                seed:  maphash.MakeSeed(),
        }, nil
 }
@@ -98,12 +97,8 @@
 // DecodeLabels transforms eBPF map key bytes into a list of label values
 // according to configuration
 func (s *Set) DecodeLabels(in []byte, labels []config.Label) ([]string, error) {
-       cacheKey := maphash.Bytes(s.seed, in)
-       if cached, ok := s.cache[cacheKey]; ok {
-               // Make sure that there is no collision here
-               if bytes.Equal(cached.in, in) {
-                       return cached.out, nil
-               }
+       if cached, ok := s.cache[string(in)]; ok {
+               return cached.out, nil
        }

        values, err := s.decodeLabels(in, labels)
@@ -111,7 +106,7 @@
                return nil, err
        }

-       s.cache[cacheKey] = cacheValue{in: in, out: values}
+       s.cache[string(in)] = cacheValue{in: in, out: values}

        return values, nil
 }
```

bobrik:
Added another commit thanks to @uhthomas:

### Use string([]byte) as a cache key in decoder

Go has a special case for `string([]byte)` that doesn't allocate:

* https://github.com/golang/go/commit/f5f5a8b6209f8

Before and after:

```
BenchmarkCache/cached-10   132218238   9.292 ns/op       0 B/op       0 allocs/op
BenchmarkCache/cached-10   212642139   5.694 ns/op       0 B/op       0 allocs/op
```

uhthomas (approved):

bobrik:
I forgot about concurrent access and there was no test to save me: #345.

bobrik:
Another missed edge case: #346.

PR #346, opened 2024-02-06
Partition decoder cache by name

bobrik:
Sometimes the same input means different things for different `[]config.Label`, but our cache implementation was not partitioning cache. With this commit we partition the cache by incoming `name` to account for that, with `name` coming from the map name, which is unique across all loaded programs.

### DIFF
--- a/decoder/decoder_test.go
+++ b/decoder/decoder_test.go
@@ -121,13 +121,13 @@ func TestDecodeLabels(t *testing.T) {
 		},
 	}
 
-	for _, c := range cases {
+	for i, c := range cases {
 		s, err := NewSet()
 		if err != nil {
 			t.Fatal(err)
 		}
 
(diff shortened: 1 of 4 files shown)


decoder/decoder_test.go
```diff
@@ -207,6 +207,66 @@ func TestConcurrency(t *testing.T) {
...
+	single, err := s.DecodeLabels(in, "one", one)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(single) != 1 {
+		t.Errorf("Expected one u64 from %#v, got %#v", one, single)
+	}
+
+	double, err := s.DecodeLabels(in, "two", two)
```
bobrik, line 260: Having the same `name` here produces the confusion.

bobrik:
We more complex `kallsyms` and `fanotify` decoders in the `Set`, so I decided against making a new `Set` per map.

PR #363, opened 2024-03-23
Split metrics (cached) and tracing (uncached) label decoding

bobrik:
In tracing there is no concern with cardinality and one can have PID as a label. With caching enabled, this means that we grow the cache map indefinitely, leaking memory until OOM stops us. Let's disable caching for tracing instead.

### DIFF
--- a/decoder/decoder.go
+++ b/decoder/decoder.go
@@ -85,9 +85,10 @@ func (s *Set) decode(in []byte, label config.Label) ([]byte, error) {
 	return result, nil
 }
 
-// DecodeLabels transforms eBPF map key bytes into a list of label values
-// according to configuration (different label sets require different names)
-func (s *Set) DecodeLabels(in []byte, name string, labels []config.Label) ([]string, error) {
+// DecodeLabelsForMetrics transforms eBPF map key bytes into a list of label values
+// according to configuration (different label sets require different names).
+// This decoder method variant does caching and is suitable for metrics.
+func (s *Set) DecodeLabelsForMetrics(in []byte, name string, labels []config.Label) ([]string, error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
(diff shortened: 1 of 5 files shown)


bobrik:
Perhaps an LRU with 1k entries per config is a better option here.

bobrik:
With LRU there's overhead in terms of CPU for the metrics case:

```
ivan@vm:~/projects/ebpf_exporter$ ~/go/bin/benchcmp /tmp/old.txt /tmp/new.txt
benchcmp is deprecated in favor of benchstat: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat
benchmark                   old ns/op     new ns/op     delta
BenchmarkCache/direct-8     382           395           +3.30%
BenchmarkCache/cached-8     20.9          35.7          +70.30%

benchmark                   old allocs     new allocs     delta
BenchmarkCache/direct-8     13             13             +0.00%
BenchmarkCache/cached-8     0              0              +0.00%

benchmark                   old bytes     new bytes     delta
BenchmarkCache/direct-8     216           216           +0.00%
BenchmarkCache/cached-8     0             0             +0.00%
```

There is no overhead for metrics with separate methods.

For tracing we have a set of nearly unique labels that are not worth caching:

```yaml
        - name: trace_id
          size: 16
          decoders:
            - name: hex
        - name: parent_span_id
          size: 8
          decoders:
            - name: hex
        - name: span_id
          size: 8
          decoders:
            - name: hex
        - name: span_monotonic_timestamp_ns
          size: 8
          decoders:
            - name: uint
        - name: span_duration_ns
          size: 8
          decoders:
            - name: uint
```

Given that we cache the whole input with unique and non-unique fields, there's zero chance of actually caching anything for tracing. There doesn't seem to be any point in LRU then.

dqminh (approved):

Issue #371, opened 2024-04-03
examples/ext4dist.bpf.c seems to have too few entries allowed in the latency map 


siebenmann:
examples/ext4dist.bpf.c appears to specify the maximum capacity of the ``ext4_latency_seconds`` hash map as ``MAX_LATENCY_SLOT + 1`` entries. However, the code seems like it will try to store up to that many separate bucket entries for each separate file operation (so ``MAX_LATENCY_SLOT`` entries for reads, and again for writes, and so on). This seems like it doesn't fit in the maximum entries, and I believe it explains some oddities I saw in actual use that went away when I updated this to multiply the limit by the number of operations. Since I'm inexperienced with eBPF programming, I may be missing something here.

If this is a real bug, it seems likely that this lack of accounting for multiple operation types may exist in other examples as well. In some cases, such as biolatency.bpf.c, this may be normally camouflaged by the fact that few people have anywhere near 255 disks, so the extra space for unused disks compensates for the lack of accounting for the various operations for most people (with modest numbers of disks).

bobrik:
You are right. Would you like to make a PR to fix this?

cc @alebsys

siebenmann:
Unfortunately I feel I don't know what the right fix for this would be. Define a new enum element at the end, called maybe ``F_MAX``, and multiply by that in the ``max_entries`` definition?

(Fixing things like biolatency.bpf.c looks even trickier since the possible req_ops values are non-contiguous, with ``REQ_OP_LAST`` set to 36 but only 14 REQ_OP_* values currently defined.)

bobrik:
> Define a new enum element at the end, called maybe F_MAX, and multiply by that in the max_entries definition?

That's what I would do.

For biolatency you can add a new `define` in ebpf code itself with the number of ops we actually use.

PR #373, opened 2024-04-04
Correctly size latency BPF hash maps in multi-operation latency programs

siebenmann:
Several BPF examples (ext4dist, xfsdist, and biolatency) accumulate latency histograms for multiple operations but size their BPF hash maps used for the latency histogram buckets as if they were only accumulating a single operation, reported as issue #371. These two commits update all three examples to correctly size their BPF hash maps. For ext4dist and xfsdist this is simple because the number of operations is set by the example's code. For biolatency the different operations come from the kernel, so we opt to be relatively conservative and size by the number of currently known operations as of 6.9-rc2. We could use the larger ``REQ_OP_LAST`` from include/linux/blk_types.h if desired; this last changed in 2022.

In ext4dist and xfsdist, the new ``F_MAX`` constant is deliberately set off from the other constants and does not have a comma after it since nothing should be added after it.

I could not find any other examples that seemed to have this pattern of under-sized BPF hash maps for latency, so I believe this will fully resolve the issues discovered in #371.

### DIFF
--- a/examples/biolatency.bpf.c
+++ b/examples/biolatency.bpf.c
@@ -6,6 +6,10 @@
 // Max number of disks we expect to see on the host
 #define MAX_DISKS 255
 
+// Max number of different request ops we expect to see in use.
+// As of 6.9-rc2, Linux defines 14 REQ_OP_* in include/linux/blk_types.h
+#define MAX_REQ_OPS 14
+
 // 27 buckets for latency, max range is 33.6s .. 67.1s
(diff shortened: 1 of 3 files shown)


bobrik:
CI is unhappy about formatting (I think there's a tab where a space should be).

siebenmann:
I've updated things and hopefully it will work and I did all the magic properly. At least I think the clang-format check passes.

bobrik:
Thanks!

PR #374, opened 2024-04-04
examples/biolatency: provide names for more request operations

siebenmann:
Define names for the relatively stable request operations beyond read, write, and flush. Discard and drv_in are routinely seen on live systems, and the others are possible so they might as well be defined. There are no names defined for zone operations because they were renumbered around v6.8-rc1.


bobrik:
Thanks!

PR #375, opened 2024-04-05
examples/biolatency: correct the name of operation 35

siebenmann:
Bio request operation 35 is REQ_OP_DRV_OUT, and should be called 'drv_out' in the label. Unfortunately when I added it in a previous commit I duplicated the name of the previous entry so it was mis-labeled as 'drv_in'. This is especially bad as it could lead to label duplications in visible Prometheus metrics.

### DIFF
--- a/examples/biolatency.yaml
+++ b/examples/biolatency.yaml
@@ -27,7 +27,7 @@ metrics:
                 5: secure_erase
                 9: write_zeroes
                 34: drv_in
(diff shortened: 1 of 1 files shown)


bobrik:
> This is especially bad as it could lead to label duplications in visible Prometheus metrics.

It wouldn't, we accumulate timeseries with the label sets into one.

PR #376, opened 2024-04-07
Add padding label config key

### LINKED ISSUE #369: Option to easily discard bytes explicitely without label
As described in the documentation, padding bytes in the counter key struct need to be considered when decoding the values.
The documentation suggests just to include the padding space in the previous value.

In case the structure is not as simple, it can get tricky and I need to define "reserved" fields. Those end up as label unnecessarily.

So, in order to avoid that, with a "discard" decoder I could explicitely discard such padding bytes.

I'd be happy to provide a patch, if you like the idea. Regarding the naming of  such a decoder I'm open to suggestions.

Edit: It seems better to just add a simple "discard" or "skip" config flag to the label.

aibor:
Padding bytes in the map key struct need to be considered in the config of the label. This is done by choosing a larger size for the label's value. For most cases this works fine.

This commit adds a new config key for labels that allows to specify padding bytes explicitly so they can be skipped by decoders.

Closes #369

### DIFF
--- a/.vscode/config-schema.yaml
+++ b/.vscode/config-schema.yaml
@@ -111,6 +111,8 @@ definitions:
           type: string
         size:
           type: number
+        padding:
+          type: number
         decoders:
           type: array
           items:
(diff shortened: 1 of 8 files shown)


bobrik (approved):

bobrik:
Thanks!

PR #379, opened 2024-04-08
Fix tracing screenshot path

bobrik:


bobrik:
Ideally we'd add a linter for this, but the linter is broken for anchors:

* https://github.com/tcort/markdown-link-check/issues/304

bobrik:
See #413.

PR #244, opened 2023-08-04
Use fanotify to monitor cgroups if available

bobrik:

PR #457, opened 2024-08-01
Learning ebpf_exporter by adding examples/cgroup-rstat-flushing

netoptimizer:
This PR implements example cgroup-rstat-flushing

- For analyzing cgroup rstat flushing behavior and lock contention issue observed in production

Thanks to for @bobrik for reviewing as I need to learn how to use ebpf_exporter

- For other to learn, I've kept some commits that show mistakes and followup commits that fixes these

### DIFF
--- a/examples/cgroup-rstat-flushing.bpf.c
+++ b/examples/cgroup-rstat-flushing.bpf.c
@@ -86,6 +86,12 @@ struct flush_key_t {
     u32 level;
 } __attribute__((packed));
 
+/* This provide easy way to enable per cgroup flush recording
+ *  Default this is disabled, as it generates too much data.
+ *  For troubleshooting purposes this can be enabled on some servers.
+ */
+#undef CONFIG_TRACK_PER_CGROUP_FLUSH
+#ifdef CONFIG_TRACK_PER_CGROUP_FLUSH
 struct {
     __uint(type, BPF_MAP_TYPE_PERCPU_HASH);
     __uint(max_entries, 10000);
@@ -99,6 +105,7 @@ struct {
     __type(key, struct flush_key_t);
     __type(value, u64);
 } cgroup_rstat_flush_nanoseconds_count SEC(".maps");
+#endif /* CONFIG_TRACK_PER_CGROUP_FLUSH */
 
 /* Complex key for encoding lock properties */
 struct lock_key_t {
@@ -386,12 +393,14 @@ int BPF_PROG(cgroup_rstat_flush_locked_exit, struct cgroup *cgrp)
      *   rate(_seconds_sum[1m]) / rate(_seconds_count[1m])
      */
 
+#ifdef CONFIG_TRACK_PER_CGROUP_FLUSH
     /* Per cgroup record average latency via nanoseconds_count and nanoseconds_sum */
     struct flush_key_t flush_key;
     flush_key.cgrp_id = cgroup_id(cgrp);
     flush_key.level = cgrp->level;
     increment_map_nosync(&cgroup_rstat_flush_nanoseconds_sum, &flush_key, delta_ns);
     increment_map_nosync(&cgroup_rstat_flush_nanoseconds_count, &flush_key, 1);
+#endif
 
     return 0;
 }
--- a/examples/cgroup-rstat-flushing.bpf.c
+++ b/examples/cgroup-rstat-flushing.bpf.c
@@ -80,6 +80,26 @@ struct {
     __type(value, u64);
 } start_flush SEC(".maps");
 
+
+struct flush_key_t {
+	u64 cgrp_id;
+	u32 level;
+} __attribute__((packed));
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_HASH);
+    __uint(max_entries, 10000);
+    __type(key, struct flush_key_t);
+    __type(value, u64);
+} cgroup_rstat_flush_seconds_sum SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_HASH);
+    __uint(max_entries, 10000);
+    __type(key, struct flush_key_t);
+    __type(value, u64);
+} cgroup_rstat_flush_seconds_count SEC(".maps");
+
 /* Complex key for encoding lock properties */
 struct lock_key_t {
     u8 contended;
@@ -257,8 +277,33 @@ int BPF_PROG(cgroup_rstat_flush_locked_exit, struct cgroup *cgrp)
 
     struct hist_key_t key;
     increment_exp2_histogram_nosync(&cgroup_rstat_flush_latency_seconds, key, delta, MAX_LATENCY_SLOT);
+    /*
+     * ebpf_exporter will also have:
+     *  ebpf_exporter_cgroup_rstat_flush_latency_seconds_sum and
+     *  ebpf_exporter_cgroup_rstat_flush_latency_seconds_count
+     *
+     * Unfortunately _sum isn't getting updated.
+     *
+     * The Prometheus idea behind having _seconds_count and _seconds_sum
+     * =================================================================
+     * The _seconds_count is number of observed flush calls that have been made, so
+     * rate(_seconds_count[1m]) in query returns per-second rate of flushes.
+     *
+     * The _seconds_sum is the sum of the delta values, so rate(flush_seconds_sum[1m])
+     * is the amount of time spent for flushes per second.
+     *
+     * Divide these two expressions to get the average latency over the last minute.
+     * The full expression for average latency would be:
+     *   rate(_seconds_sum[1m]) / rate(_seconds_count[1m])
+     */
 
-    // IDEA: cgroup_rstat_flush_seconds_total with label cgroup_id ?
+    /* Per cgroup record average latency via _seconds_count and _seconds_sum */
+    struct flush_key_t flush_key;
+    flush_key.cgrp_id = cgroup_id(cgrp);
+    flush_key.level = cgrp->level;
+    increment_map_nosync(&cgroup_rstat_flush_seconds_sum, &flush_key, delta);
+    increment_map_nosync(&cgroup_rstat_flush_seconds_count, &flush_key, 1);
+
     return 0;
 }
 
--- a/examples/cgroup-rstat-flushing.bpf.c
+++ b/examples/cgroup-rstat-flushing.bpf.c
@@ -83,8 +83,8 @@ int BPF_PROG(rstat_locked, struct cgroup *cgrp, int cpu, bool contended)
 		(*cnt)++;
 	}
 
-	/* What cgrp level is interesting, but I didn't manage to encode it in
-	 * above counters.  As contended case is the most interesting, have
+	/* What cgroup level is interesting, but I didn't manage to encode it
+	 * in above counters.  As contended case is the most interesting, have
 	 * level counter for contended.
 	 */
 	if (contended) {
--- a/examples/cgroup-rstat-flushing.bpf.c
+++ b/examples/cgroup-rstat-flushing.bpf.c
@@ -90,14 +90,14 @@ struct {
     __uint(max_entries, 10000);
     __type(key, struct flush_key_t);
     __type(value, u64);
-} cgroup_rstat_flush_seconds_sum SEC(".maps");
+} cgroup_rstat_flush_nanoseconds_sum SEC(".maps");
 
 struct {
     __uint(type, BPF_MAP_TYPE_PERCPU_HASH);
     __uint(max_entries, 10000);
     __type(key, struct flush_key_t);
     __type(value, u64);
-} cgroup_rstat_flush_seconds_count SEC(".maps");
+} cgroup_rstat_flush_nanoseconds_count SEC(".maps");
 
 /* Complex key for encoding lock properties */
 struct lock_key_t {
@@ -264,14 +264,15 @@ int BPF_PROG(cgroup_rstat_flush_locked_exit, struct cgroup *cgrp)
     u64 now = bpf_ktime_get_ns();
     struct start_time_key_t key_ts;
     u64 *start_flush_ts;
-    u64 delta;
+    u64 delta, delta_ns;
 
     key_ts.pid_tgid = bpf_get_current_pid_tgid();
     key_ts.cgrp_id = cgroup_id(cgrp);
     read_array_ptr(&start_flush, &key_ts, start_flush_ts);
     // TODO: validate LRU lookup success
     /* Flush time latency */
-    delta = (now - *start_flush_ts) / 100; /* 0.1 usec */
+    delta_ns = now - *start_flush_ts;
+    delta = delta_ns / 100; /* 0.1 usec */
     *start_flush_ts = 0;
 
     struct hist_key_t key;
@@ -296,12 +297,12 @@ int BPF_PROG(cgroup_rstat_flush_locked_exit, struct cgroup *cgrp)
      *   rate(_seconds_sum[1m]) / rate(_seconds_count[1m])
      */
 
-    /* Per cgroup record average latency via _seconds_count and _seconds_sum */
+    /* Per cgroup record average latency via nanoseconds_count and nanoseconds_sum */
     struct flush_key_t flush_key;
     flush_key.cgrp_id = cgroup_id(cgrp);
     flush_key.level = cgrp->level;
-    increment_map_nosync(&cgroup_rstat_flush_seconds_sum, &flush_key, delta);
-    increment_map_nosync(&cgroup_rstat_flush_seconds_count, &flush_key, 1);
+    increment_map_nosync(&cgroup_rstat_flush_nanoseconds_sum, &flush_key, delta_ns);
+    increment_map_nosync(&cgroup_rstat_flush_nanoseconds_count, &flush_key, 1);
 
     return 0;
 }
--- a/examples/cgroup-rstat-flushing.bpf.c
+++ b/examples/cgroup-rstat-flushing.bpf.c
@@ -13,6 +13,12 @@
 
 #define MAX_CGROUP_LEVELS 5
 
+/* From: linux/include/linux/cgroup.h */
+static inline u64 cgroup_id(const struct cgroup *cgrp)
+{
+    return cgrp->kn->id;
+}
+
 struct {
     __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
     __uint(max_entries, MAX_CGROUP_LEVELS + 1);
@@ -54,6 +60,25 @@ struct {
     __type(key, u64);
     __type(value, u64);
 } start_hold SEC(".maps");
+
+struct {
+    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+    __uint(max_entries, MAX_LATENCY_SLOT + 1);
+    __type(key, struct hist_key_t);
+    __type(value, u64);
+} cgroup_rstat_flush_latency_seconds SEC(".maps");
+
+struct start_time_key_t {
+    u64 pid_tgid;
+    u64 cgrp_id;
+};
+
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __uint(max_entries, 1024 * 128);
+    __type(key, struct start_time_key_t);
+    __type(value, u64);
+} start_flush SEC(".maps");
 
 /* Complex key for encoding lock properties */
 struct lock_key_t {
@@ -116,7 +141,6 @@ int BPF_PROG(rstat_locked, struct cgroup *cgrp, int cpu, bool contended)
 
     /* Lock hold time start */
     bpf_map_update_elem(&start_hold, &pid, &now, BPF_ANY);
-    // TODO: Should we ignore yield and measure flush time instead?
 
     /* Lock contended event happened prior to obtaining this lock.
      * Get back start "wait" timestamp that was recorded.
@@ -128,6 +152,7 @@ int BPF_PROG(rstat_locked, struct cgroup *cgrp, int cpu, bool contended)
 
         read_array_ptr(&start_wait, &pid, start_wait_ts);
         // TODO: validate LRU lookup success
+	/* Lock wait time */
         delta = (now - *start_wait_ts) / 100; /* 0.1 usec */
 
         increment_exp2_histogram_nosync(&cgroup_rstat_lock_wait_seconds, key, delta, MAX_LATENCY_SLOT);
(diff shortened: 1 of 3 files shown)


examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+
+ Concern: The lock rates will vary (a lot) and aggregating this as an average
+ will not capture spikes. Especially given Prometheus capture intervals only
+ happens every 53 seconds.
+
+ Q: will lock "type=yield" be a good idea?
+
+*/
+
+SEC("tp_btf/cgroup_rstat_locked")
```
bobrik, line 72: This is too new for the Ubuntu we run in CI, so you need to add an exclude to `CONFIGS_TO_IGNORE_IN_CHECK` in the top level `Makefile`.

examples/cgroup-rstat-flushing.yaml
```diff
@@ -0,0 +1,31 @@
...
+          decoders: # contended boolean converted to 0 and 1
+            - name: uint
+    - name: cgroup_rstat_lock_contended
+      help: Lock contention counters per cgroup level
+      labels:
+        - name: level
+          size: 4
+          decoders:
+            - name: uint
+
```
bobrik, line 31: Let's drop this empty line.

examples/cgroup-rstat-flushing.yaml
```diff
@@ -0,0 +1,31 @@
...
+          decoders:
+            - name: uint
+    - name: cgroup_rstat_locked_total
+      help: Total number of times rstat lock was obtainted and contended state
+      labels:
+        - name: contended
+          size: 4
+          decoders: # contended boolean converted to 0 and 1
+            - name: uint
+    - name: cgroup_rstat_locked_yield
```
bobrik, line 17: `cgroup_rstat_locked_yields_total`, see https://prometheus.io/docs/practices/naming/

examples/cgroup-rstat-flushing.yaml
```diff
@@ -0,0 +1,31 @@
...
+      help: Number of times rstat lock was obtainted again after yield and contended state
+      labels:
+        - name: contended
+          size: 4
+          decoders: # contended boolean converted to 0 and 1
+            - name: uint
+    - name: cgroup_rstat_lock_contended
```
bobrik, line 24: `cgroup_rstat_lock_contentions_total`, I think.

examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+ */
+
+#include <vmlinux.h>
+#include <bpf/bpf_tracing.h>
+#include "maps.bpf.h"
+
+#define MAX_CGRP_LEVELS	5
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
```
bobrik, line 17: The tabs are making clang-format unhappy in CI.

examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+
+ Mitigation workaround: Store counters per cgroup "level" (level=0 is root).
+ This will allow us to separate root-cgroup flushes from cadvisor walking all
+ cgroup levels.
+
+ */
+SEC("fentry/cgroup_rstat_flush_locked")
+int BPF_PROG(cgroup_rstat_flush_locked, struct cgroup *cgrp)
+{
+	u64 level_key = cgrp->level;
```
bobrik, line 129: Should this be u32 to match `__type(key, u32)`?

examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+
+*/
+
+SEC("tp_btf/cgroup_rstat_locked")
+int BPF_PROG(rstat_locked, struct cgroup *cgrp, int cpu, bool contended)
+{
+	u32 key = contended;
+	u64 *cnt;
+
+	read_array_ptr(&cgroup_rstat_locked_total, &key, cnt);
```
bobrik, line 78: Shouldn't `increment_map_nosync` work here as well?

examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+ *  - https://github.com/xdp-project/xdp-project/blob/master/areas/latency/cgroup_rstat_tracepoint.bt
+ *
+ * Depends on tracepoints added in kernel v6.10
+ */
+
+#include <vmlinux.h>
+#include <bpf/bpf_tracing.h>
+#include "maps.bpf.h"
+
+#define MAX_CGRP_LEVELS	5
```
bobrik, line 14: nit: `MAX_CGROUP_LEVELS`

examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,139 @@
...
+
+	read_array_ptr(&cgroup_rstat_locked_total, &key, cnt);
+	(*cnt)++;
+
+	if (cpu >= 0) {
+		read_array_ptr(&cgroup_rstat_locked_yield, &key, cnt);
+		(*cnt)++;
+	}
+
+	/* What cgrp level is interesting, but I didn't manage to encode it in
```
bobrik, line 86: nit: `cgrp` -> `cgroup`

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: IMHO then BPF  restricted-C code should be written with same tabs style as kernel code.
Lets change for formatting requirements, please!

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: Makes sense.
I'll do that.

examples/cgroup-rstat-flushing.yaml
netoptimizer: done

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: This is a good catch - thanks!

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: sure

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: Sure to nit.
I was hoping to get your input to the contents of the comment:
```
	/* What cgrp level is interesting, but I didn't manage to encode it in
	 * above counters.  As contended case is the most interesting, have
	 * level counter for contended.
	 */
```
When coding BPF, I usually do **one** map lookup to get a stats `struct`, that I will update with multiple stats.
Here I end-up doing 3 map lookups for single counters.

When I stated coding I though I could represent something like:
```
ebpf_exporter_cgroup_rstat_locked_total{contended="0" level="0"} 6558
ebpf_exporter_cgroup_rstat_locked_total{contended="1" level="0"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="0" level="1"} 34
ebpf_exporter_cgroup_rstat_locked_total{contended="1" level="1"} 123
ebpf_exporter_cgroup_rstat_locked_total{contended="0" level="2"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="1" level="2"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="0" level="3"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="1" level="3"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="0" level="4"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="1" level="4"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="0" level="5"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="1" level="5"} 0
```
Question: But ...
1. I cannot figure out how to get ebpf_exporter to generate this?
2. I don't know if these counters makes sense for Prometheus ?

examples/cgroup-rstat-flushing.yaml
netoptimizer: Thanks for the [link](https://prometheus.io/docs/practices/naming/) , as a Prometheus newbie I didn't realize that the __suffix_ had a meaning.
I now realize that general strucure is _library_name_unit_suffix_.

Jakub recommended: [Prometheus: Up & Running, 2nd Edition](https://www.oreilly.com/library/view/prometheus-up/9781098131135/)

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: Code just evolved from doing some operations on the returned pointer "object".
So, now this is a pointer to a simple counter I can use `increment_map_nosync` as I do in other places.

examples/cgroup-rstat-flushing.bpf.c
bobrik: > I don't know if these counters makes sense for Prometheus ?

They do

> I cannot figure out how to get ebpf_exporter to generate this?

Each line here is a separate key in a single map called `ebpf_exporter_cgroup_rstat_locked_total`:

```
ebpf_exporter_cgroup_rstat_locked_total{contended="0", level="0"} 6558
ebpf_exporter_cgroup_rstat_locked_total{contended="1", level="0"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="0", level="1"} 34
ebpf_exporter_cgroup_rstat_locked_total{contended="1", level="1"} 123
ebpf_exporter_cgroup_rstat_locked_total{contended="0", level="2"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="1", level="2"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="0", level="3"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="1", level="3"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="0", level="4"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="1", level="4"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="0", level="5"} 0
ebpf_exporter_cgroup_rstat_locked_total{contended="1", level="5"} 0
```

The mapping is effectively:

* `{contended="0", level="0"}` -> `6558`
* `{contended="1", level="0"}` -> `0`
* `{contended="0", level="1"}` -> `34`
* ...

To encode each key, you turn the raw bytes of the key into the values with the help of decoders.

Working backwards from what you expect on the output, you want something like this in the config:

```yaml
    - name: cgroup_rstat_locked_total
      help: Total number of times rstat lock was obtainted and contended state
      labels:
        - name: contended
          size: 1
          decoders:
            - name: uint
        - name: level
           size: 1
           decoders:
            - name: uint
```

On the kernel side you'd want something like this to represent the key:

```c
struct key_t {
    u8 contended;
    u8 level;
};
```

The names here do not matter, only sizes do (u8 corresponds to `size: 1` in the config).

Given that your keys are complex structs and not small numbers, you cannot store these in an array, so you need a hashmap:

```
struct {
    __uint(type, BPF_MAP_TYPE_PERCPU_HASH);
    __uint(max_entries, 128);
    __type(key, struct key_t);
    __type(value, u64);
} cgroup_rstat_locked_total SEC(".maps");
```

To actually increment a value, you need to construct the key and to the increment in the map:

```c
// construct however you need to
struct key_t key = {};
key.contended = 1;
key.level = 2;

increment_map_nosync(&cgroup_rstat_locked_total, &key, 1);
```

Here's an example with a complex key:

* https://github.com/cloudflare/ebpf_exporter/blob/master/examples/pci.yaml
* https://github.com/cloudflare/ebpf_exporter/blob/master/examples/pci.bpf.c

Hope this makes sense.

netoptimizer: These suggestions worked for encoding more information ("labels" in Prometheus terms). 
I've coded it up and pushed as part of this patchset/PR.
(p.s. I've also encoded "yield" state)

examples/cgroup-rstat-flushing.bpf.c
bobrik: Let's leave it for another issue. For now let's just make clang-format happy.

examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,211 @@
...
+	u64 now = bpf_ktime_get_ns();
+	u64 pid = bpf_get_current_pid_tgid();
+	u64 *start_hold_ts;
+	struct hist_key_t key;
+	u64 delta_usec;
+
+	/* Lock hold time */
+	read_array_ptr(&start_hold, &pid, start_hold_ts);
+	// TODO: validate LRU lookup success
+	delta_usec = (now - *start_hold_ts) / 100; /* 0.1 usec */
```
bobrik, line 154: This isn't really `usec` anymore. Any reason not to keep it in ns?

examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,211 @@
...
+	u32 level = cgrp->level;
+
+	if (level > MAX_CGROUP_LEVELS)
+		level = MAX_CGROUP_LEVELS;
+
+	if (cpu >= 0)
+		lock_key.yield = 1;
+
+	lock_key.contended = contended;
+	lock_key.level = (level & 0xFF);
```
bobrik, line 113: What does this do?

examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,211 @@
...
+ happens every 53 seconds.
+
+*/
+SEC("tp_btf/cgroup_rstat_locked")
+int BPF_PROG(rstat_locked, struct cgroup *cgrp, int cpu, bool contended)
+{
+	u64 now = bpf_ktime_get_ns();
+	u64 pid = bpf_get_current_pid_tgid();
+	struct lock_key_t lock_key = { 0 };
+	u32 level = cgrp->level;
```
bobrik, line 104: You can operate on `lock_key.level` directly, no need for a separate variable.

examples/cgroup-rstat-flushing.bpf.c
```diff
@@ -0,0 +1,211 @@
...
+	increment_map_nosync(&cgroup_rstat_locked_total, &lock_key, 1);
+
+	/* Lock hold time start */
+	bpf_map_update_elem(&start_hold, &pid, &now, BPF_ANY);
+	// TODO: Should we ignore yield and measure flush time instead?
+
+	/* Lock contended event happened prior to obtaining this lock.
+	 * Get back start "wait" timestamp that was recorded.
+	 */
+	if (contended) {
```
bobrik, line 124: This can be an early return.

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: Good point with variable `usec` naming, will fix.

Keeping it in nanosec will use more buckets, and I'm trying to reduce how much data we feed Prometheus.
Maybe I don't need to reduce size?

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: True for current code, but I might add more code later.
And I think the code is more readable IMHO, as it is more clear that everything under `if (contended) {` is handling the contended case.

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: Below I'm using `level` and bounding it to `MAX_CGROUP_LEVELS`.
I prefer to keep level type `u32` in this bounding code, and do the assignment later to `lock_key.level` as it is type `u16`.

examples/cgroup-rstat-flushing.yaml
netoptimizer: This map was removed after improvement to encode this as labels.

examples/cgroup-rstat-flushing.bpf.c
netoptimizer: This projects `.clang-format` file states:
`# Minimal format loosely matching the kernel style.`
IMHO it doesn't match kernel style at all...

As it contains:
```
IndentWidth: 4
TabWidth: 4
UseTab: Never
```
Looking at kernel source in `.clang-format` this should have been:
```
IndentWidth: 8
TabWidth: 8
UseTab: Always
```

I will respect that this project requires its own special clang-format rule and obey to these, but I still think they are wrong.

examples/cgroup-rstat-flushing.bpf.c
bobrik: I personally would stick to a usec level resolution. Not sure if you care about buckets below 1us.

Whatever works for you is fine, as long as the naming doesn't contradict the contents. We can always change it later.

examples/cgroup-rstat-flushing.bpf.c
bobrik: I would normally just cast it explicitly: `lock_key.level = (u16) level`, which should have the same effect.

bobrik:
Let me know when it's ready to go. It would be helpful if you attached the exact metrics that you see locally with this running and capturing some events.

netoptimizer:
Running this locally overnight, I see an issue with cgroup decoding: e.g `cgroup="unknown_cgroup_id:2196814`
Longer output:
```
ebpf_exporter_cgroup_rstat_flush_seconds_sum{cgroup="unknown_cgroup_id:2196562",level="2"} 224
ebpf_exporter_cgroup_rstat_flush_seconds_sum{cgroup="unknown_cgroup_id:2196598",level="2"} 305
ebpf_exporter_cgroup_rstat_flush_seconds_sum{cgroup="unknown_cgroup_id:2196670",level="2"} 172
ebpf_exporter_cgroup_rstat_flush_seconds_sum{cgroup="unknown_cgroup_id:2196706",level="2"} 167
ebpf_exporter_cgroup_rstat_flush_seconds_sum{cgroup="unknown_cgroup_id:2196742",level="2"} 210
[...]
curl -s localhost:9435/metrics | grep unknown_cgroup_id | grep _sum | wc -l
6043
```
Something on my system seems to be creating cgroups for a short period and ebpf_exporter is picking these up
-  I assume what happens is that:
- when decoding these for (Prometheus) output, they are gone and we get these "unknown_cgroup_id"
- (for completeness: I think `pmcd` / `pmlogger` is causing this, but it is irrelevant)

I'm want to change the code such that this per-cgroup stats cannot explode like this
- considering using a LRU map with a limit (e.g. 1024)
- as we are primarily interested in the most active cgroups
- but as `cadvisor` walks all cgroups we have to be careful limit is always above cgroups in system to avoid LRU trashing corner-cases

netoptimizer:
Regarding above "unknown_cgroup_id:nnn" issue (Cc @bobrik )

Can the _cgroup_ **decoder** be instructed to exclude unknown cgroups in the output?
- or only make a single "unknown_cgroup" output record?

bobrik:
> Something on my system seems to be creating cgroups for a short period and ebpf_exporter is picking these up

Does ebpf_exporter say anything about `fanotify` (#244) on startup? Even short lived cgroups should be resolved when fanotify is available.

> Can the cgroup decoder be instructed to exclude unknown cgroups in the output?
> or only make a single "unknown_cgroup" output record?

Not in the current state, but it seems like a good idea: #460.

Another option, which might not be possible: check for cgroup age and if it's less than some threshold (say 10s), set cgroup_id to zero to do the collapsing manually.

netoptimizer:
> > Something on my system seems to be creating cgroups for a short period and ebpf_exporter is picking these up
> 
> Does ebpf_exporter say anything about `fanotify` (#244) on startup? Even short lived cgroups should be resolved when fanotify is available.
> 

I don't think `fanotify` is avail/configured on my testlab.

> > Can the cgroup decoder be instructed to exclude unknown cgroups in the output?
> > or only make a single "unknown_cgroup" output record?
> 
> Not in the current state, but it seems like a good idea: #460.

Awesome that you created a ticket for this :smiley:

bobrik:
> I don't think fanotify is avail/configured on my testlab.

It's a matter of having v6.6 or newer: https://github.com/torvalds/linux/commit/0ce7c12e88cf. There isn't anything to configure. As long as ebpf_exporter doesn't say anything about fanotify at startup, it is using it.

netoptimizer:
Some example output running this in "localhost" mode
- e.g. on port 9435
-  extracted via command: `curl -s localhost:9435/metrics`

```
# HELP ebpf_exporter_cgroup_rstat_locked_total Times rstat lock was obtainted with state for cgroup level, contended and yield
# TYPE ebpf_exporter_cgroup_rstat_locked_total counter
ebpf_exporter_cgroup_rstat_locked_total{contended="0",level="0",yield="0"} 24281
ebpf_exporter_cgroup_rstat_locked_total{contended="0",level="0",yield="1"} 7351
ebpf_exporter_cgroup_rstat_locked_total{contended="0",level="1",yield="0"} 14533
ebpf_exporter_cgroup_rstat_locked_total{contended="0",level="1",yield="1"} 39
ebpf_exporter_cgroup_rstat_locked_total{contended="0",level="2",yield="0"} 175181
ebpf_exporter_cgroup_rstat_locked_total{contended="0",level="2",yield="1"} 89
ebpf_exporter_cgroup_rstat_locked_total{contended="0",level="3",yield="0"} 272801
ebpf_exporter_cgroup_rstat_locked_total{contended="0",level="3",yield="1"} 119
ebpf_exporter_cgroup_rstat_locked_total{contended="0",level="4",yield="0"} 1805
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="0",yield="0"} 86951
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="0",yield="1"} 1.2791901e+07
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="1",yield="0"} 73
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="1",yield="1"} 3864
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="2",yield="0"} 722
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="2",yield="1"} 22764
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="3",yield="0"} 1077
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="3",yield="1"} 34344
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="4",yield="0"} 10
ebpf_exporter_cgroup_rstat_locked_total{contended="1",level="4",yield="1"} 399
```

bobrik (approved):

PR #527, opened 2025-02-26
add a lru cache to regexp decoder

dqminh:
Often, regexp decoder can run repeatedly on the same input. An common example is matching cgroup path in a chain like so:

```
- name: cgroup
- name: regexp
  regexps:
  - ^.*(system.slice).*$
```

In these cases, the overhead of regexp matching can often be noticable. Since the input does not vary too often, we can add a cache here to skip the overhead of regex matching. This is implemented as an LRU cache to prevent the cache from growing infinitely, especially in cases where we have many short-live cgroups. The previous example can be written aborts

```
- name: cgroup
- name: regexp
  lru_cache_size: 1024
  regexps:
  - ^.*(system.slice).*$
```


decoder/regexp.go
```diff
@@ -6,22 +6,32 @@ import (
...
 func (r *Regexp) Decode(in []byte, conf config.Decoder) ([]byte, error) {
 	if conf.Regexps == nil {
 		return nil, errors.New("no regexps defined in config")
 	}
+	inputStr := string(in)
 
 	if r.cache == nil {
 		r.cache = map[string]*regexp.Regexp{}
 	}
+	if conf.LruCacheSize > 0 && r.outputCache == nil {
```
bobrik, line 28: There's already a cache for label sets, shouldn't it cover this too and at a higher level?

https://github.com/cloudflare/ebpf_exporter/blob/b62525ba2dc2780521da95f3c41bf5d9cb266f47/decoder/decoder.go#L96-L100

dqminh: @bobrik Yes, you are right. The LRU cache does not make sense. I think this should have been a negative cache instead, because the majority of overhead comes from matching input that would have produced ErrSkipLabelSet.
I've adjusted the implementation as such.

bobrik: I don't think we need to limit the size here, as we don't do it for positive lookups. The most likely scenario is that you see the whole possible set of values on every scrape, so having an LRU that's one element too small will make it evict on every call and never hit anything. We don't have metrics, so it's easy to get this wrong and get somewhat higher CPU usage as an outcome.

Given that the decoding process is supposed to be deterministic (and we fully cache successes already), I think we need to cache errors as well and in the same place:

https://github.com/cloudflare/ebpf_exporter/blob/b62525ba2dc2780521da95f3c41bf5d9cb266f47/decoder/decoder.go#L108-L111

This will be both faster (as we do while set instead of one decoder kind) and universal.

bobrik: Thinking about it some more: we skip things we don't want in metrics and they can in fact be higher cardinality than the thing we want to retain. Perhaps we do want an LRU cache after all, but it still makes sense to have it at the global level.

Not sure what to do about the sizing. Maybe log a message if too many errors are produced and hit rate is bad?

dqminh: Yes, i think we can lift the cache to be the same level as global, and make this a default behavior with configurable size. The question is whether we want multiple negative LRU cache, or only a single one. In practice, i think only regexp currently skips labels with potentially high cardinalities, so i think a single global LRU cache should be ok. It's simpler to configure too.

Regarding hit rates, I guess we can produce a prometheus metrics for this, and it's user responsibility to determine the desirable size by looking at the metrics and also load of the exporter.

bobrik: SGTM

PR #530, opened 2025-03-02
Remove redaundant check in maps.bpf.h

tahajahangir:
Remove the `max_bucket` check from `_increment_ex2_histogram`, the check is also done in `_increment_histogram`


bobrik:
Thanks!

PR #531, opened 2025-03-03
Feature: synchronize cgroup id with ebpf program via shared map

dqminh:
For some usecases, it's useful for ebpf programs to filter their metrics based on cgroups. It's generally hard to ebpf program to be able to filter cgroup at runtime. This make it easier by allowing ebpf exporter to update known interesting cgroup id at runtime via a shared BPF map.

This requires cgroup fanotify support in kernel to work.


bobrik:
LGTM, except for the linter errors.

PR #543, opened 2025-04-28
Exporter to measure cephfs performance

niko-lay:
Added exporter to measure cephfs performance, similar to ext4dist
Exposed metrics are 
 ```
 READ,
 WRITE,
 MKDIR,
 UNLINK,
 FSYNC
```


bobrik:
Thanks!

PR #565, opened 2025-08-04
Support for BPF_MAP_CGRP_STORAGE type

dqminh:
BPF_MAP_TYPE_CGRP_STORAGE (https://docs.kernel.org/bpf/map_cgrp_storage.html) provides a fast solution to attach some data to cgroup. This should be better than getting data from a centralize hashmap, so provide some support for that in cgroup id map.


netoptimizer:
For completeness, there is a followup PR#566

PR #634, opened 2026-04-28
examples: Add unix-accept-latency

jsitnicki:
ebpf_exporter ships examples covering TCP accept queue latency but no equivalent for AF_UNIX. Listener-side accept queue stalls on busy UNIX stream sockets are invisible to the existing examples.

Add an example that measures the time a connection spends on the listener's accept queue and exports it as an exp2 histogram, unix_accept_latency_seconds, labeled by listener path. Track three common server sockets out of the box: /run/dbus/system_bus_socket, /run/systemd/journal/stdout, and /var/run/mysqld/mysqld.sock. Operators are expected to edit the path list to match the host.

Attach an fentry on security_unix_stream_connect() to stamp the server-side newsk in BPF_MAP_TYPE_SK_STORAGE when the listener's unix_sock->addr->name[0].sun_path matches one of the monitored entries. On unix_accept() fexit, look up the stamp on the accepted sk, compute the delta, and feed it into the histogram. Storing the matched path index alongside the timestamp avoids re-walking unix_sock->addr on the accept side; sk_storage presence already proves the socket is monitored.

Latencies are bucketed in milliseconds from 1 ms to 1024 ms. Samples above the top edge are clamped into the last bucket.

### DIFF
--- a/examples/unix-accept-latency.bpf.c
+++ b/examples/unix-accept-latency.bpf.c
@@ -1,0 +1,149 @@
+// Measure connect-to-accept latency for monitored UNIX stream sockets.
+//
+// Stores an enqueue timestamp in sk_storage on the server-side socket
+// at security_unix_stream_connect(), then reads it back at unix_accept()
+// return to compute the time spent on the listener's accept queue.
+
+#include <vmlinux.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+#include "bits.bpf.h"
+#include "maps.bpf.h"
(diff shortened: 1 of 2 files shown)


bobrik:
CI is unhappy, presumably because of kernel version mismatch. Can you take a look?

bobrik:
It was actually https://github.com/cloudflare/ebpf_exporter/pull/635, so you probably just need to rebase.

care44cubs-max (commented):
Yes

PR #655, opened 2026-08-21
Allow loading of local module symbols

arges:
This expands kallsyms scanning to allow symbols that are from modules. In addition, if we have duplicate symbols prefer the first-loaded name. This also adds unit tests as needed.


bobrik:
I merged #656 to fix the linter, you need to rebase. LGTM otherwise.

PR #658, opened 2026-08-25
add a size limit to the decoder label cache

Eninspace:
The label cache in `Set.cache` is a plain map that is only ever written to:

```go
cache[string(in)] = values
```

It is keyed by raw eBPF map key bytes. For maps keyed by something with high
cardinality it grows without bound: the kernel side is bounded
(`examples/tcp-retransmit.bpf.c` uses `BPF_MAP_TYPE_LRU_HASH` with
`max_entries`), but every key the kernel LRU evicts leaves a permanent entry
behind in the decoder cache. On a long running exporter this shows up as steady
heap growth, with most of the live objects sitting on the cache insert line.

This adds a size limit for the label cache, customizable with the flag
`config.cache-size`, mirroring `config.skip-cache-size`. It defaults to zero,
which keeps the current unbounded behaviour.


bobrik:
Thanks!

